PR #122, opened 2024-04-17
Ceyhun/keys

ceyhunsen:
# Description

This PR introduces keys module, which handles reading or creating private and public keys depending on what the user is supplying.

## Linked Issues

- None

## Testing

- cargo test
- cargo run

## Docs

Module and functions are documented using /// blocks.

### DIFF
--- a/core/src/keys.rs
+++ b/core/src/keys.rs
@@ -1,0 +1,39 @@
+//! This module reads or creates private and public keys depending on what the
+//! user is supplying.
+
+use crate::constants::NUM_VERIFIERS;
+use bitcoin::XOnlyPublicKey;
+use crypto_bigint::rand_core::OsRng;
+use secp256k1::{All, Secp256k1, SecretKey};
+use std::env;
+
+/// Gets private and public keys, either from a user supplied source or creates
+/// random pairs. User can supply keys as files using PUBLIC_KEYS and
+/// PRIVATE_KEYS environment variables.
+pub fn get_keys(secp: Secp256k1<All>, rng: &mut OsRng) -> (Vec<SecretKey>, Vec<XOnlyPublicKey>) {
+    match read_from_file() {
+        Some(ret) => return ret,
+        None => (),
+    }
+
+    create(secp, rng)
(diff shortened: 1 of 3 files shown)


ozankaymak (commented):
Might need a example .env file.

ceyhunsen:
Current problems:

1. Do we read private keys from a single file or multiple files?
2. Do we generate public keys from private keys.
3. Errors should be moved to `core/src/errors.rs`

Reading both private and public keys complicate code. Generating public keys will simplify code.

ozankaymak:
Nice PR, well commented. I wish I were leaving comments and instructions like you.

ozankaymak (approved):
Maybe creating a single file for a single run of generating sks and pks is better, @ekrembal?

core/src/bin/key_generator.rs
```diff
@@ -0,0 +1,68 @@
...
+
+/// Directory to put generated key files.
+const DIRECTORY: &str = "configs";
+
+/// Key file prefix.
+const PREFIX: &str = "keys";
+
+fn main() {
+    let (all_sks, all_xonly_pks) = generate_keypair();
+    println!("Generated private keys: {:#?}", all_sks.clone());
```
ekrembal, line 23: Lets use tracing here

core/src/bin/key_generator.rs
```diff
@@ -0,0 +1,68 @@
...
+use core::panic;
+use crypto_bigint::rand_core::OsRng;
+use secp256k1::SecretKey;
+use std::{
+    fs::{self, File},
+    io::Write,
+};
+
+/// Directory to put generated key files.
+const DIRECTORY: &str = "configs";
```
ekrembal, line 16: We can also get this directory from env variable

ekrembal (changes requested):

core/src/bin/key_generator.rs
ekrembal: Makes sense since this is a one time use binary only for testing.

PR #130, opened 2024-04-22
Database Refactoring

### LINKED ISSUE #128: Mock DB with file read & write
### Issue Description

We need to modify databases so that they read from a file, and every time an update function is called, they should read from the file, deserialize the structure, update it, serialize it again, and write it back to the file. Additionally, making the database functions asynchronous (while maintaining synchronous code internally) will be beneficial for actual database scenarios.

### Tasks

Parts of the task:
- [ ] Make DB struct Serializable
- [ ] Write read and write logic
- [ ] Change DB functions

### Deadline of the Issue

24 April

ceyhunsen:
# Description

This PR introduces DB operations over real files.

## Linked Issues

- Closes #128

## Testing

- Unit tests are added for new modules.
- `cargo test`

## Docs

Interfaces are documented in code. No external documentation is required.


ceyhunsen:
This PR mostly focused on `core/src/db/common.rs` and `core/src/db/text.rs`. Other modified files are mostly consists renaming. Therefore review is mostly needed on these 2 files.

Important point in this PR is the synchronization stuff. Please read comments in `core/src/db/common.rs` to understand the solution. Review is mostly needed in that section.

There is a problem: Running whole program second time results failure. Because operator and verifier are created in the first run, second run returns an error called `AlreadyInitialized`. This might not be a problem right now if that part of the code is under active refactoring. If not, we need to fix this.

ekrembal (approved):

PR #133, opened 2024-04-24
DB functions: Update return values and make them async

### LINKED ISSUE #132: DB functions: Update return values and make them async
### Issue Description

DB functions currently does not returns any kind of error, if database is not ready or not have requested information. They should return a `Result` type for error checking. This will be useful in the future, when a real DBMS is replaced with mock DB.

Also, DB functions should use tokio's `async` feature.

### Tasks

Parts of the task:

- [x] Convert DB function's return type to `Result`
- [x] Refactor other parts of the code to handle the new `Result` return type
- [x] Add `async` functionality to DB functions
- [x] Add new tests if needed for the new `async` functionality

### Deadline of the Issue

26.04.2024

ceyhunsen:
# Description

This PR introduces 2 key changes:

* DB functions returns `Result` instead of panicking in case of error
* DB functions support Tokio's async functionality

## Linked Issues

- Closes #132 

## Testing

* `cargo test`
* `cargo run`

## Docs

Not much to be documented. It is only a small refactor for existing code.

### DIFF
--- a/core/src/errors.rs
+++ b/core/src/errors.rs
@@ -102,6 +102,9 @@ pub enum BridgeError {
     /// Given key pair is invalid and new pairs can't be generated randomly
     #[error("InvalidKeyPair")]
     InvalidKeyPair,
+    /// Database could not handle request and returned an error
+    #[error("DatabaseError")]
+    DatabaseError(std::io::Error),
 }
 
 impl From<secp256k1::Error> for BridgeError {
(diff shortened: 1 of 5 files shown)


core/src/errors.rs
```diff
@@ -159,3 +162,9 @@ impl From<MerkleBlockError> for BridgeError {
         BridgeError::MerkleBlockError(err)
     }
 }
+
+impl From<std::io::Error> for BridgeError {
```
ekrembal, line 166: Let's not convert every std error to Database Error

ekrembal (changes requested):

core/src/errors.rs
ceyhunsen: Done

PR #165, opened 2024-05-22
Mock Bitcoin RPC interface

ceyhunsen:
# Description

This PR introduces a mock Bitcoin RPC interface for testing. This is achieved with the [Bitcoin Simulator](https://github.com/Bitcoin-Wildlife-Sanctuary/bitcoin-simulator), a third party Bitcoin simulator.

## Docs


ceyhunsen:
This PR should be chained with future PR's. We didn't do much work here. So, we can discard changes here.

PR #187, opened 2024-06-28
Use global SECP

### LINKED ISSUE #182: Use global secp context
### Proposal Description

From https://github.com/ark-network/clArk/blob/master/ark-lib/src/util.rs#L7-L10

```rust
lazy_static::lazy_static! {
	/// Global secp context.
	pub static ref SECP: secp256k1::Secp256k1<secp256k1::All> = secp256k1::Secp256k1::new();
}
```

ekrembal:
# Description
Uses global secp context instead of passing, 

## Linked Issues

- Closes #182

### DIFF
--- a/core/src/utils.rs
+++ b/core/src/utils.rs
@@ -9,13 +9,31 @@ use bitcoin::taproot::LeafVersion;
 use bitcoin::taproot::TaprootSpendInfo;
 use bitcoin::Amount;
 use bitcoin::ScriptBuf;
+use bitcoin::XOnlyPublicKey;
 use hex;
 use sha2::{Digest, Sha256};
 use std::borrow::BorrowMut;
+use std::str::FromStr;
 
 lazy_static::lazy_static! {
-	/// Global secp context.
-	pub static ref SECP: secp256k1::Secp256k1<secp256k1::All> = secp256k1::Secp256k1::new();
+    /// Global secp context.
+    pub static ref SECP: secp256k1::Secp256k1<secp256k1::All> = secp256k1::Secp256k1::new();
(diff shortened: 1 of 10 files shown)


ceyhunsen:
Second flow test fails on my machine:

```
---- test_flow_2 stdout ----
thread 'test_flow_2' panicked at core/tests/flow.rs:213:88:
called `Result::unwrap()` on an `Err` value: ControlBlockError
```

core/src/utils.rs
```diff
@@ -9,9 +9,32 @@ use bitcoin::taproot::LeafVersion;
...
+    /// Global secp context.
+    pub static ref SECP: secp256k1::Secp256k1<secp256k1::All> = secp256k1::Secp256k1::new();
+}
+
+lazy_static::lazy_static! {
+    /// This is an unspendable pubkey.
+    ///
+    /// See https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#constructing-and-spending-taproot-outputs
+    pub static ref UNSPENDABLE_PUBKEY: bitcoin::secp256k1::PublicKey =
+        "93c7378d96518a75448821c4f7c8f4bae7ce60f804d03d1f0628dd5dd0f5de51".parse().unwrap();
```
ceyhunsen, line 28: Unwrap is here very unlikely to cause problems in the future because it's nearly impossible to fail. But I'd say we should ditch every `.unwrap()` call eventually and return error messages in a user friendly manner.

ceyhunsen (approved):
I mean it's alright. Like... Overrated as baklava in my opinion.

PR #191, opened 2024-07-01
Get rid of script builder struct

ekrembal:
# Description

We actually don't need a seperate script builder

## Testing
None


ceyhunsen (approved):
No struct (or class for the OOP people) based functions?! This reminds me of my C days.

LGTM.

PR #216, opened 2024-08-13
xonly from pks

### LINKED ISSUE #213: Clean code dup in musig2 tests
### Issue Description

https://github.com/chainwayxyz/clementine/blob/ekrem/new-architecture/core/src/musig2.rs#L116

### Tasks

Parts of the task:
- [ ] Add another function to create XOnlyPubkey from a vector of PublicKeys
- [ ] Fix type conversions

### Deadline of the Issue

14 Aug

lemonpartee:
Closes #213


core/src/musig2.rs
```diff
@@ -125,10 +125,32 @@ mod tests {
...
 
+    trait FromPublicKeys {
+        fn from_pks(
+            pks: Vec<PublicKey>,
+            tweak: Option<[u8; 32]>,
+        ) -> (musig2::secp256k1::PublicKey, XOnlyPublicKey);
+    }
+
+    impl FromPublicKeys for XOnlyPublicKey {
+        fn from_pks(
```
ekrembal, line 140: Do you think we need to describe this from_pks is using musig2?

RomanHodulak: Good point, I renamed it to `from_musig2_pks`. Let me know if you have other ideas for the name.

Although on another look, I think since it is a private trait inside `musig2` module it might be self-explanatory.

lemonpartee: Good point, I renamed it to `FromMuSig2Pks::from_musig2_pks`. Let me know if you have other ideas for the name.

On another look, I think since it is a private trait inside `musig2` module it might be redundant. On the other hand, it is true that there are other kinds of public keys this could be confused with.

ekrembal (approved):

PR #222, opened 2024-08-15
Update DB for New Architecture

### LINKED ISSUE #212: Fill db functions
### Issue Description

Similar to #201 we need to fill unimplemented db function. We can have the same schema for verifiers and operators

### Tasks

_No response_

### Deadline of the Issue

16.08.2024

ozankaymak:
This PR introduces the changes on the database operations for both Postgre and Rust.
Closes #212.


ozankaymak:
Added some checks for OutPoints. Not adding checks for Bech32 format for now since Taproot addresses might be encoded differently under different Networks. You can see [this thread](https://stackoverflow.com/questions/21683680/regex-to-match-bitcoin-addresses).
We still need to decide on where we are going to use OutPoints, TxOuts or UTXOs. @ekrembal?

ekrembal:
Let's skip the regex for taproot addresses.
lets make id's auto increment
rename deposit_request_outpoint  -> deposit_outpoint

ozankaymak:
We need a mechanism for the operators to handle their kickoff UTXOs efficiently. Currently, they will use a chain of transactions to generate kickoffs, where each transaction has the first k (variable) outputs that can be used as kickoff UTXOs, and then the change output for future kickoff UTXO generation, and finally, a anyone-can-pay output for CPFP. I believe this could be improved. In the meantime, we need to handle it in the database part as well.

ozankaymak:
All functionalities are implemented in this PR; however, some things still need to be implemented:
- Operators' logic for finding the kickoff_utxo they used for a withdrawal they provided (chaining of the funding transactions)
- Generic column name use when implementing FromRow trait
- Instead of for loop, handle the multiple queries better

ekrembal (approved):
We need tests for db functions but i will open another issue for that

PR #232, opened 2024-08-23
Deposit test

lemonpartee:


core/tests/flow.rs
```diff
@@ -2,336 +2,219 @@
...
+    let secret_key = secp256k1::SecretKey::new(&mut secp256k1::rand::thread_rng());
+
+    let user = User::new(
+        rpc.clone(),
+        config.verifiers_public_keys.clone(),
+        secret_key,
+        config.clone(),
+    );
+
+    let deposit_outpoints: Vec<_> = evm_addresses
```
ekrembal, line 88: We should generate a single deposit_outpoint for this test, we will just test the flow of a single deposit

core/tests/flow.rs
```diff
@@ -2,336 +2,219 @@
...
+
+    rpc.mine_blocks(18).unwrap();
+
+    // for every verifier, we call new_deposit
+    // aggregate nonces
+    let mut agg_nonces = Vec::new();
+
+    for (i, (client, _)) in verifiers.iter().enumerate() {
+        let musig_pub_nonces = client
+            .verifier_new_deposit_rpc(
```
ekrembal, line 108: All the verifiers should get the same deposit request, so that they will sign a correct thing

ekrembal (changes requested):

ekrembal:
We still need to aggreagete partial signatures, #233 This should help, This flow test should end with move_tx going onchain

ekrembal (approved):

Issue #252, opened 2024-08-28
Test flow for withdrawal


ekrembal:

ekrembal:
Closed with #253

PR #253, opened 2024-08-28
Full withdrawal flow

### LINKED ISSUE #221: Operator withdrawal function

ekrembal:
# Description
Withdrawal happy part test.
Closes #221 
More tests needed but related to #252


core/src/database/wrapper.rs
```diff
@@ -24,7 +24,7 @@ pub struct EVMAddressDB(pub EVMAddress);
 #[derive(Serialize, Deserialize)]
 pub struct TxidDB(pub Txid);
 
-#[derive(Serialize, Deserialize)]
+#[derive(Serialize, Deserialize, sqlx::FromRow)]
```
lemonpartee, line 27: This seems like a column type, `FromRow` should be implemented for a single item tuple like this `(SignatureDb,)`.

ekrembal: cc: @ozankaymak

ozankaymak: I mean it works so maybe later we can change this after we are done with all this? Because I think in general we need to change the database module.

PR #262, opened 2024-08-30
Separation of mock and real RPC tests

ceyhunsen:
# Description

This PR separates mock and real rpc tests. This will help identify which test failed easily.

## Testing

Tested in PR itself.


ceyhunsen:
I think using Ubicloud results cache saving to fail. Is this possible? If it is, can we do anything to solve it? @cetin

ozgune:
Hi @ceyhunsen - I'm one of the founders at Ubicloud. If you think this is related to the Ubicloud cache, please feel free to drop us a note at support@ubicloud.com. We'd be happy to take a look.

cetin (dismissed):

cetin (approved):

PR #264, opened 2024-08-30
Unify server binaries

ekrembal:


ceyhunsen:
Adding how to use this features to readme can help. Do we want to outsiders to use these binaries?

ceyhunsen (commented):
Works as expected. But code looks weird.

In this context we don't need small performance benefits while starting servers. Because of this, we can actually use CLI flags passed to binary itself. This will result:

- One binary to every use case
- Binaries won't get big because actual application code is the smallest part of our binaries
- We won't be able to limit user to do one thing and one thing only; Users are free to run which server they want

With CLI flags, this can be the usage (which is clearer for me):

```bash
cargo run --verifier-server --operator-server --aggregator-server
```

ozankaymak:
IMO @ceyhunsen is right; however, README.md should itself completely change because a lot of things changed since the design was changed. I think it deserves to be a seperate issue. @ekrembal ?

ozankaymak (approved):

ekrembal:
Yes, will update README but not first priority. I think using a single binary is not that good and this implementation is simpler

ceyhunsen (approved):

PR #268, opened 2024-09-03
Use the same ports in binary, 0 in tests

### LINKED ISSUE #269: Make test config 7 verifiers 3 operators
### Issue Description

For better understand the real world case

### Tasks

_No response_

### Deadline of the Issue

-

ekrembal:
Closes #269


ozankaymak (dismissed):

core/src/servers.rs
```diff
@@ -111,6 +111,11 @@ pub async fn create_aggregator_server(
     Ok((client, handle, addr))
 }
 
+fn is_test_env() -> bool {
+    // if thread name is not main then it is a test
+    thread::current().name().unwrap_or_default() != "main"
```
ceyhunsen, line 116: Maybe use `if cfg!()`?

```rust
if cfg!(test) {
    true
} else {
   false
}
```

ekrembal: it didn't work

ceyhunsen: Does this also not work?

```rust
#[cfg(test)]
true
#[cfg(not(test)]
false
```

ceyhunsen: Reported as not working again

ozankaymak (approved):

ceyhunsen (dismissed):

PR #271, opened 2024-09-03
Update mock RPC

ceyhunsen:
# Description

Updates mock RPC version


ceyhunsen:
It must be fixed. There is an obvious problem. One more version update will be on the way.

ekrembal (approved):

PR #285, opened 2024-09-09
Ozan/withdrawal fee

ozankaymak:
This PR adds a basic withdrawal profitability check for operators.


.github/actions/service-action/action.yml
```diff
@@ -79,4 +80,7 @@ runs:
       db_name = "clementine"
       citrea_rpc_url = "http://159.89.214.47/"
       bridge_contract_address = "3100000000000000000000000000000000000002"
+      citrea_rpc_url = "http://127.0.0.1:12345"
+      citrea_rpc_user = "admin"
+      citrea_rpc_password = "admin"
```
ekrembal, line 85: We don't need user and password

core/src/operator.rs
```diff
@@ -305,24 +305,32 @@ where
...
+        if withdrawal_amount > self.config.bridge_amount_sats {
+            Ok(false)
+        } else {
+            // Calculate net profit after the withdrawal
+            let net_profit = self.config.bridge_amount_sats - withdrawal_amount;
+
+            // Check if the operator withdrawal fee is set and calculate profitability
+            match self.config.operator_withdrawal_fee_sats {
+                Some(fee) => Ok(fee < net_profit),
+                None => Err(BridgeError::OperatorWithdrawalFeeNotSet),
```
ekrembal, line 319: We should give error on initialization not in when it's called

core/src/operator.rs
```diff
@@ -305,24 +305,32 @@ where
...
-        withdrawal_idx: usize,
+        _withdrawal_idx: usize,
         user_sig: schnorr::Signature,
         input_utxo: UTXO,
         output_txout: TxOut,
     ) -> Result<Option<Txid>, BridgeError> {
         // TODO: check that withdrawal_idx has the input_utxo.outpoint
 
-        if !self.is_profitable(withdrawal_idx).await? {
+        if !self.is_profitable(output_txout.value.to_sat()).await? {
```
ekrembal, line 333: lets add input's dust value here as well. Let's not force users to create only dust values, they can create 1 BTC and request 11 BTC

core/tests/data/test_config.toml
```diff
@@ -46,3 +47,6 @@ db_port = 5432
 db_user = "clementine"
 db_password = ""
 db_name = "clementine"
+citrea_rpc_url = "http://127.0.0.1:12345"
+citrea_rpc_user = "admin"
+citrea_rpc_password = "admin"
```
ekrembal, line 52: We don't need this

ekrembal (changes requested):

PR #290, opened 2024-09-10
Withdrawal fixes

ekrembal:


.github/actions/service-action/action.yml
ozankaymak, line 1: Why remove?

ozankaymak (dismissed):

.github/actions/service-action/action.yml
ceyhunsen: Default config file's db values are correct for the CI/CD env. Therefore, no reason to create an overwrite config file.

ozankaymak (approved):

ceyhunsen (approved):

PR #307, opened 2024-09-16
Unit tests for where the code coverage is not present

ceyhunsen:
# Description

This PR adds unit tests to various modules that lacks code coverage.

## Testing

Comment out `#[tracing::instrument]` bits in the code and then run:

```sh
cargo llvm-cov --html
```

Then, open `target/llvm-cov/html/index.html` in a browser.

## Code Coverage Improvements

Numbers stated here could be misleading. Looks like macros sometimes cause wrong coverage reports. Results after disabling `#[tracing::instrument]`s:

Before:

image

After:

image

## Skipped Code Portions

- `operator.rs`: Skipped testing code that checks and does some operations if Citrea client is present
- General database/encode/decode operations: Some operations can throw errors. Most of the cases, the error is caused by the module itself or user. It is not possible mimic a module failure or corrupt user data, mid-test.
- ? are ignored: They are mostly caused by the user and handled by the Rust itself. No need to check if error is not generated by the Clementine itself. It is a binary's responsible to handle these checks and print a helpful information.
- `tracing::debug!` like macros are reported as not covered. They are ignored. But it is noted that they are lowering code coverage percentage.

### DIFF
--- a/core/src/actor.rs
+++ b/core/src/actor.rs
@@ -59,7 +59,7 @@ impl Actor {
     }
 
     #[tracing::instrument(skip(self), err(level = tracing::Level::ERROR), ret(level = tracing::Level::TRACE))]
-    pub fn sign_taproot_script_spend_tx_new(
+    pub fn sign_taproot_script_spend_tx(
         &self,
         tx: &mut TxHandler,
         txin_index: usize,
--- a/core/src/actor.rs
+++ b/core/src/actor.rs
@@ -5,7 +5,7 @@ use bitcoin::sighash::SighashCache;
 use bitcoin::taproot::LeafVersion;
 use bitcoin::{
     hashes::Hash,
-    secp256k1::{ecdsa, schnorr, Keypair, Message, SecretKey, XOnlyPublicKey},
+    secp256k1::{schnorr, Keypair, Message, SecretKey, XOnlyPublicKey},
     Address, TapSighash, TapTweakHash,
 };
 use bitcoin::{TapLeafHash, TapNodeHash, TapSighashType, TxOut};
@@ -13,7 +13,7 @@ use bitcoin::{TapLeafHash, TapNodeHash, TapSighashType, TxOut};
 #[derive(Debug, Clone)]
 pub struct Actor {
     pub keypair: Keypair,
-    secret_key: SecretKey,
+    _secret_key: SecretKey,
     pub xonly_public_key: XOnlyPublicKey,
     pub public_key: secp256k1::PublicKey,
     pub address: Address,
@@ -28,7 +28,7 @@ impl Actor {
 
         Actor {
             keypair,
-            secret_key: keypair.secret_key(),
+            _secret_key: keypair.secret_key(),
             xonly_public_key: xonly,
             public_key: keypair.public_key(),
             address,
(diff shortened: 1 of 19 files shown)


core/src/extended_rpc.rs
```diff
@@ -271,15 +153,6 @@ where
         self.client.get_raw_transaction(txid, block_hash)
     }
 
-    #[tracing::instrument(skip(self), err(level = tracing::Level::ERROR), ret(level = tracing::Level::TRACE))]
-    pub fn get_transaction(
```
ozankaymak, line 275: Why did we remove these functions? These functions do not need test since we did not implement them.

core/src/operator.rs
```diff
@@ -329,14 +329,6 @@ where
         }
     }
 
-    /// Checks if utxo is valid, spendable by operator and not spent
-    /// Saves the utxo to the db
-    #[tracing::instrument(skip(self), err(level = tracing::Level::ERROR), ret(level = tracing::Level::TRACE))]
-    async fn set_funding_utxo(&self, funding_utxo: UTXO) -> Result<(), BridgeError> {
-        self.db.set_funding_utxo(None, funding_utxo).await?;
```
ozankaymak, line 336: I am not sure if we should delete this.

core/src/extended_rpc.rs
ceyhunsen: They weren't used. @ekrembal said we should delete the functions we don't use.

core/src/operator.rs
ceyhunsen: Same as above. I can re-add them if needed.

ozankaymak (dismissed):

core/src/operator.rs
ekrembal: We should not delete this, indeed we should write tests that uses this

ekrembal (changes requested):

core/src/operator.rs
ceyhunsen: Re added with a unit test.

ozankaymak (approved):

core/src/actor.rs
```diff
@@ -58,54 +58,6 @@ impl Actor {
...
-                &tx.scripts[txin_index][script_index],
-                LeafVersion::TapScript,
-            ),
-            bitcoin::sighash::TapSighashType::Default,
-        )?;
-        Ok(sig_hash)
-    }
-
     #[tracing::instrument(skip(self), err(level = tracing::Level::ERROR), ret(level = tracing::Level::TRACE))]
     pub fn sign_taproot_script_spend_tx_new(
```
ekrembal, line 62: Can we also rename this sign_taproot_script_spend_tx

core/src/database/common.rs
```diff
@@ -1177,7 +1161,7 @@ mod tests {
                 .await
                 .unwrap()
                 .unwrap();
-            tracing::info!("unused_utxo: {:?}", unused_utxo);
+            println!("unused_utxo: {:?}", unused_utxo);
```
ekrembal, line 1164: why println here?

core/src/user.rs
```diff
@@ -44,30 +44,9 @@ where
         }
     }
 
-    #[tracing::instrument(skip(self), err(level = tracing::Level::ERROR), ret(level = tracing::Level::TRACE))]
-    pub fn deposit_tx(
```
ekrembal, line 48: Can we also re-add this, and change the tests to use this maybe?

ekrembal (changes requested):

core/src/user.rs
ceyhunsen: 2 of the return values were known by the caller and not modified inside of the function. So, they are removed.

core/src/transaction_builder.rs
```diff
@@ -23,468 +21,426 @@ pub struct TxHandler {
 
 pub type CreateAddressOutputs = (Address, TaprootSpendInfo);
 
-#[derive(Debug, Clone)]
-pub struct TransactionBuilder {
-    _verifiers_xonly_pks: Vec<XOnlyPublicKey>,
-    _verifiers_pks: Vec<PublicKey>,
-    _network: Network,
-}
-
```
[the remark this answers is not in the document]
ozankaymak: No problem, lgtm

ekrembal (approved):

ozankaymak (approved):

PR #312, opened 2024-09-23
cargo: Remove debug flags and add release flags.

### LINKED ISSUE #162: Compiling Takes Long Time and Uses Excessive Amount of Memory
### Issue Description

Compiling any target, like tests or a specific binary, takes a very long time and consumes high memory. Sometimes, my machine with 16 GB RAM fails to compile (OS kills both processes because of the OOM) because of high memory usage (100% of the 16 GB). This happens when VS Code's Rust Analyzer and Cargo working together. This is a typical scenario and a typical environment for a developer and this should not be acceptable for a cli program.

I guess this issue probably caused by the multiple binary targets. In that case, a different approach should be proposed to solve this issue, for the sake of my (or possibly others) sanity. I don't have any proposal right now, though.

It also might be caused by VS Code's configuration. In that case, we might put a default local configuration (`.vscode`) to repo itself. Don't think this is the reason though.

Let's fix this, pls.

### Tasks

Parts of the task:

- [ ] Make sure what makes both VS Code and Cargo consumes excessive amount of memory and time
- [ ] Propose a solution
- [ ] Implement that solution

### Deadline of the Issue

2024

ceyhunsen:
# Description

This PR modifies Cargo flags to achieve faster compilation times in tests and faster binaries in release mode.

Using default configuration for dev/test builds lowered compiling times significantly (on a Linux machine with 12th gen i5):

- Optimization level 3 (previous): 85.3s | 239M
- Optimization level z (for size): 57.6s | 150M
- Optimization level 0 (Cargo default): 21.6s | 180M

For debug/test builds, lower compilation times are king.

For release mode, some optimizations are enabled to lower binary size and improve execution speed. But resulted with increased compilation times.

## Linked Issues

- Closes #162 

## Testing

Build times are compared with Cargo's `--timings` flag. Binary sizes are compared with `du`.

All existing tests are checked if they are running correctly.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,16 +29,10 @@ sqlx = { version = "0.7.4", default-features = false }
 bitcoin-mock-rpc = { git = "https://github.com/chainwayxyz/bitcoin-mock-rpc", tag = "v0.0.11" }
 musig2 = { version = "0.0.11", features = ["serde"] }
 
-# Always optimize; building and running the guest takes much longer without optimization.
-[profile.dev]
-opt-level = 3
-
-[profile.dev.build-override]
-opt-level = 3
-
 [profile.release]
-debug = 1
 lto = true
-
-[profile.release.build-override]
-opt-level = 3
+strip = true
+codegen-units = 1
+# TODO: Abort on panic can be enabled to save binary space. But needs a proper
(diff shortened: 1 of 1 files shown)


ekrembal (approved):
Cool 🔥

PR #328, opened 2024-10-01
Simplify server RPC layers

### LINKED ISSUE #313: Remove redundant RPC endpoint layer
### Proposal Description

Operator, verifier and aggregator has RPC trait implementation that calls the actual functions. Maybe that extra layer can be removed and actual function called directly.

ceyhunsen:
# Description

This PR removes redundant RPC layer for servers.

## Linked Issues

- Closes #313


ceyhunsen:
Currently, there is a possibility to switch gRPC from json-RPC. So, this work is probably useless. Can be opened if plans are changed.

Issue #325, opened 2024-09-30
Research database transaction details and refactor code


ceyhunsen:
### Proposal Description

Research if database transactions are prone to multi threaded read/write operations. If not, refactor some parts of the code to not use transaction to gain some performance. Example: https://github.com/chainwayxyz/clementine/pull/309/commits/786d1b45fc15abc880c567f94b1e069ce2e5bee5

ceyhunsen:
PostgreSQL transactions appears to be hygienic (https://www.postgresql.org/docs/current/tutorial-transactions.html).

> Another important property of transactional databases is closely related to the notion of atomic updates: when multiple transactions are running concurrently, each one should not be able to see the incomplete changes made by others.

Meaning example change is OK.

ceyhunsen:
Also, select statements takes a snapshot before executing query and search for values in that snapshot, regardless using a transaction or not.

> In effect, a SELECT query sees a snapshot of the database as of the instant the query begins to run. However, SELECT does see the effects of previous updates executed within its own transaction, even though they are not yet committed. Also note that two successive SELECT commands can see different data, even though they are within a single transaction, if other transactions commit changes after the first SELECT starts and before the second SELECT starts.
>
> https://www.postgresql.org/docs/current/transaction-iso.html#XACT-READ-COMMITTED

ceyhunsen:
After some time, I come to a conclusion that this is fine. Thanks for waiting.

PR #330, opened 2024-10-02
Database crate simplification

ceyhunsen:
# Description

This PR removes redundant code and simplifies database crate:

- Verifier and operator specific database layers removed
- PostgreSQL related stuff moved to `mod.rs`
- Some of the functions are marked as private to be only used internally

## Linked Issues

- Related to #325 

## Testing

Added new tests.


core/src/bin/server.rs
```diff
@@ -20,9 +20,7 @@ async fn main() {
         config.bitcoin_rpc_password.clone(),
     );
 
-    let database = Database::new(config.clone()).await.unwrap();
-    database.init_from_schema().await.unwrap();
-    database.close().await;
+    Database::initialize_database(&config).await.unwrap();
```
ekrembal, line 23: This deletes the db, but it should not. This is dangerous

ekrembal (changes requested):

ekrembal (approved):

PR #332, opened 2024-10-03
Mock function separation

### LINKED ISSUE #181: Don't Compile Mock Interfaces With Release Binary
# Proposal Description

Currently mock interfaces sits on `src/` directory and added as module regardless which compilation target is selected (release or tests).

Mock interfaces are needed in both unit and integration tests. That's the reason why they are in `src/` directory. If it is put in `tests/` directory, unit tests in `src/` directory can't access them. But this causes a problem: Release binaries may have these mock interfaces included. This can cause unnoticed problems.

Currently, there is no common way to implement a mock interface in Rust world. If there are any better ways to achieve this, please comment.

## Problems

- Can't mark mock code with `#[cfg(test)]` because is is for unit tests and can't be used with integration tests
  - All integration tests can be moved to `src/tests` to solve this but it is not optimal
- Can't use custom feature (something like `testing`):
  - When it is disabled by default, Rust analyzer reports no function is available
  - When it is disabled by default, need to pass `--features testing` to every command invocation
  - When it is enabled by default, binaries can also use these functions
- If moved to a different crate, can't use `clementine_core` utilities because of the cyclic dependencies
- Mock code can't be hidden behind a private module because then, integration tests can't access them
- If code is put in `src` and configured out with `#[cfg(test)]`:
  - Main code path will be different for both cases: Integration tests see it as `clementine_core::` and unit tests see it as `crate::`
  - It can be solved with double conditional import but somehow `crate::` is reported to be not found

TL;DR: If a code portion is accessible to an integration test, it is also accessible to binaries. And if a code portion is put in integration test, it is not accessible to any other.

## Possible Solutions

- Have too many crates that are independent from each other:
  - `database`, `servers`, `actor`, `config`...
  - Like this, a `mock` crate can use some of these crates and itself can be a `dev-dependency` of some other crates, which are not used by it
  - Still cause cyclic dependencies if not careful
  - Way too much work is needed
- Remove `tests` dir and move integration tests inside of unit tests
  - This can help lower compilation times
  - Not very "Rust way"

ceyhunsen:
# Description

This PR separates mock functions from main code and makes it impossible to be used them in release binaries.

To do this, mock code marked with `cfg(test)`.

## Linked Issues

- Closes #181 

## Testing

added unit tests?

## Docs

have the docs been updated?


ceyhunsen:
As discussed in #181, this work is not going anywhere.

PR #371, opened 2024-12-05
Winternitz derivation

### LINKED ISSUE #334: Implement winternitz derivation
### Issue Description

We need a way to sign and generate winternitz signatures

### Tasks

Parts of the task:
- [x] research derivation paths
- [x] implement generating winternitz pubkeys in actor.rs
- [x] implement signing winternitz signatures in actor.rs

### Deadline of the Issue
25/10/2024

ceyhunsen:
# Description

Adds BitVM as dependency and uses Winternitz part to generate Winternitz one time public key.

## Linked Issues

- Closes #334 
- Related #178 
- Relatad https://github.com/chainwayxyz/risc0-to-bitvm2/pull/14


core/src/actor.rs
```diff
@@ -9,26 +9,79 @@ use bitcoin::{
...
+/// Derivation path specification for Winternitz one time public key generation.
+#[derive(Debug, Clone, Copy)]
+pub struct WinternitzDerivationPath {
+    pub message_length: u32,
+    pub log_d: u32,
+    pub index: u32,
+    pub tx_type: TxType,
+    pub operator_idx: u32,
+    pub watchtower_idx: u32,
+    pub time_tx_idx: u32,
```
ekrembal, line 30: Let's make all index, operator_idx, wathctower_idx and time_tx_idx Option, if it is None, then use 0 if not, the add 1 and use it. This way we can use zero indexed things

core/src/actor.rs
```diff
@@ -9,26 +9,79 @@ use bitcoin::{
...
+            time_tx_idx: Default::default(),
+        }
+    }
+}
 
 #[derive(Debug, Clone)]
 pub struct Actor {
     pub keypair: Keypair,
     _secret_key: SecretKey,
+    winternitz_secret_key: Option<String>,
```
ekrembal, line 64: This should be 32 bytes

core/src/actor.rs
```diff
@@ -348,4 +422,23 @@ mod tests {
             )
             .unwrap();
     }
+
+    #[tokio::test]
+    async fn derive_winternitz_pk() {
```
ekrembal, line 427: Let's also add one more test that has a hardcoded public key

ekrembal (changes requested):
Let's also add signing

core/src/actor.rs
ceyhunsen: Fixed in https://github.com/chainwayxyz/clementine/pull/371/commits/aed085261ea75ee760f5ed0ae39ccbfe90a55140.

core/src/actor.rs
ceyhunsen: Fixed in https://github.com/chainwayxyz/clementine/pull/371/commits/ec074f4c5b2bc7385c14f4a41f4e775c3fdc519f.

ekrembal (approved):

Issue #355, opened 2024-10-16
Implement Get Params


ekrembal:
# Issue Description

Depends on #334 

## Tasks

Parts of the task:
- [ ] Create get params for operator
- [x] Create get params for wathtower
- [ ] Create get patams for verifier

## get_params for watchtower

- [x] Creates Winternitz pubkeys (240 bytes) for every operator and time_tx pair
- [x] Aggregator sends this pubkeys to verifiers
- [x] Verifiers saves these to database

## Deadline of the Issue

1/11/3024

PR #374, opened 2024-12-10
Collect and set Winternitz public keys in aggregator

ceyhunsen:
# Description

Adds collecting and saving Winternitz public keys support for `aggragator::setup` RPC endpoint.

## Linked Issues

- Related to #355


core/src/aggregator.rs
```diff
@@ -73,12 +75,23 @@ impl Aggregator {
         let operator_clients =
             rpc::get_clients(operator_endpoints, ClementineOperatorClient::connect).await?;
 
+        let watchtower_endpoints =
+            config
+                .watchtower_endpoints
+                .clone()
+                .ok_or(BridgeError::ConfigError(
+                    "Couldn't find operator endpoints in config file!".to_string(),
```
ekrembal, line 83: Couldn't find **watchtower** endpoints in config file

core/src/database/common.rs
```diff
@@ -918,6 +920,56 @@ impl Database {
 
         Ok((result.0 .0, result.1 .0, result.2, receipt))
     }
+
+    /// Sets Winternitz public keys for a watchtower.
+    #[tracing::instrument(skip(self, tx, winternitz_public_keys), err(level = tracing::Level::ERROR), ret(level = tracing::Level::TRACE))]
+    pub async fn save_winternitz_public_key(
```
ekrembal, line 926: I think we might need to seperate saving these winternitz pubkeys with respect to operators but not sure.

scripts/schema.sql
```diff
@@ -130,4 +130,10 @@ create table if not exists header_chain_proofs (
     proof bytea
 );
 
+-- Verifier table of Winternitz public keys for every operator and time_tx pair
+create table if not exists winternitz_public_keys (
+    watchtower_id int primary key not null,
```
ekrembal, line 135: Let's save winternitz per operator_id not watchtower_id so that we don't have to load all winternitz pubkeys when signing

ekrembal (changes requested):

ekrembal (approved):

PR #375, opened 2024-12-12
Mock crate separation from binaries

### LINKED ISSUE #181: Don't Compile Mock Interfaces With Release Binary

ceyhunsen:
# Description

Converts all the testing functions into macros. This way, release binaries has no way to use these "testing only" functions/macros.

Let's assume we have a testing function, in `mock.rs`:

```rust
use crate::config::BridgeConfig;

pub fn test_config() -> BridgeConfig {
    BridgeConfig::default()
}
```

Unit tests will happily import this `mock.rs` and use `test_config()`. But in case of binaries and integration tests, they will complain that there is no such thing as `config::BridgeConfig` in crate root. This problem is simply solved with converting testing functions into macros. Now, every module needs to import the structures and functions used by their preference.

## Linked Issues

- Closes #181

### DIFF
--- a/core/src/actor.rs
+++ b/core/src/actor.rs
@@ -330,9 +330,11 @@ impl Actor {
 #[cfg(test)]
 mod tests {
     use super::Actor;
+    use crate::config::BridgeConfig;
+    use crate::utils::initialize_logger;
     use crate::{
         actor::WinternitzDerivationPath, builder::transaction::TxHandler,
-        mock::database::create_test_config_with_thread_name,
+        create_test_config_with_thread_name, database::Database, initialize_database,
     };
     use bitcoin::{
         absolute::Height, transaction::Version, Amount, Network, OutPoint, Transaction, TxIn, TxOut,
@@ -345,7 +347,9 @@ mod tests {
         treepp::script,
     };
     use secp256k1::{rand, Secp256k1, SecretKey};
+    use std::env;
     use std::str::FromStr;
+    use std::thread;
 
     /// Returns a valid [`TxHandler`].
     fn create_valid_mock_tx_handler(actor: &Actor) -> TxHandler {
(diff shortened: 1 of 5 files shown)


ekrembal (approved):
Great catch @ceyhunsen !

core/src/test_utils.rs
```diff
@@ -0,0 +1,316 @@
+//! # Testing Utilities
+//!
+//! This crate provides testing utilities, which are not possible to be included
```
mmtftr, line 3: you mean module?

PR #379, opened 2024-12-16
Operator BitVM assert tx Winternitz public key generation

ceyhunsen:
# Description

Adds Winternitz public key generation for operator BitVM assert txs.


core/src/operator.rs
```diff
@@ -695,6 +696,35 @@ impl Operator {
...
+    /// BitVM assert tx.
+    ///
+    /// # Returns
+    ///
+    /// - [`Vec<Vec<winternitz::PublicKey>>`]: Winternitz public keys for
+    ///   `watchtower index` row and `BitVM assert tx index` column.
+    pub fn get_winternitz_public_keys(&self) -> Result<Vec<winternitz::PublicKey>, BridgeError> {
+        let mut winternitz_pubkeys = Vec::new();
+
+        for watchtower in 0..self.config.num_watchtowers as u32 {
```
ekrembal, line 710: Every operator has time_tx times winternitz pubkeys, no need for watchtower for loop

ekrembal (changes requested):

core/src/operator.rs
ekrembal: Yes, each operator should have TimeTx amount of winternitz pubkeys for assert txs. Each assert TX has around 2000 bytes of data to sign with winternitz.

ekrembal (approved):

PR #381, opened 2024-12-17
Deposit sighash collection

ceyhunsen:
# Description

Adds sighash collection for deposit steps.


core/src/rpc/operator.rs
```diff
@@ -47,10 +47,17 @@ impl ClementineOperator for Operator {
...
-            .await;
+            .collect::<Result<_, BridgeError>>()
+            .await?;
+
+        // Generate Winternitz public keys and convert them to RPC type.
+        let winternitz_pubkeys = self.get_winternitz_public_keys()?;
+        let winternitz_pubkeys = winternitz_pubkeys
+            .into_iter()
+            .map(WinternitzPubkey::from_bitvm)
+            .collect::<Vec<_>>();
```
atacann, line 60: duplicate

core/src/actor.rs
```diff
@@ -257,13 +266,20 @@ impl Actor {
         let mut sighash_cache: SighashCache<&mut bitcoin::Transaction> =
             SighashCache::new(&mut tx_handler.tx);
 
+        let prevouts = bitcoin::sighash::Prevouts::All(&tx_handler.prevouts);
+        let leaf_hash = TapLeafHash::from_script(
+            &tx_handler
+                .scripts
+                .get(txin_index)
```
ceyhunsen, line 273: Do we need to bound checks for this? Indexing with `[idx]` is dangerous if the array is smaller than the index but in this case, is there a possibility that this would happen? If not, we can revert this for simpler code and better performance.

core/src/actor.rs
```diff
@@ -357,7 +373,7 @@ mod tests {
         let mut tx_handler = create_invalid_mock_tx_handler(actor);
 
         let prev_tx: Transaction = Transaction {
-            version: Version::TWO,
+            version: Version(3),
```
ceyhunsen, line 376: Latest rust-bitcoin does not have support for this. That's why we did it like this.

core/src/builder/sighash.rs
```diff
@@ -1,34 +1,307 @@
...
-use async_stream::stream;
-use bitcoin::TapSighash;
+use async_stream::try_stream;
 use bitcoin::{address::NetworkUnchecked, Address, Amount, OutPoint};
+use bitcoin::{TapSighash, Txid};
 use futures_core::stream::Stream;
 
-pub fn create_nofn_sighash_stream(
-    _db: Database,
+pub async fn dummy(
```
ceyhunsen, line 10: This function can be deleted: It's here because rust-analyzer won't give hints about code if it is in a macro (in this case `try_stream`).

core/tests/musig2.rs
```diff
@@ -149,6 +149,7 @@ async fn key_spend() {
         &musig_agg_xonly_pubkey_wrapped,
     )
     .unwrap();
+    rpc.mine_blocks(1).await.unwrap();
```
ceyhunsen, line 152: This and other unrelated `mine_blocks()` calls are added because current rust-bitcoin library version creates V2 transactions and Bitcoin Core will deny to use these UTXOs in V3 transactions, before getting them mined. This can be deleted after rust-bitcoin library supports V3 transactions.

core/src/rpc/operator.rs
ceyhunsen: Fixed in https://github.com/chainwayxyz/clementine/pull/381/commits/0f4b154d95d6bc2d0561be76078b4f34724fdd45.

ceyhunsen:
I suggest **holding** merge for this PR until we find the reason why grpc_flow test takes too much time. I think test is passing but it is very slow. On Github, it seems to be even more slower and Github marks this test failed because of this.

If this is not the case, there could be some race conditions.

core/src/builder/sighash.rs
```diff
@@ -1,34 +1,307 @@
...
+                    );
+                yield Actor::convert_tx_to_sighash_script_spend(
+                    &mut watchtower_challenge_txhandler,
+                    0,
+                    0,
+                )?;
+
+                let mut operator_challenge_nack_txhandler =
+                    builder::transaction::create_operator_challenge_nack_txhandler(
+                        wcp_txid,
```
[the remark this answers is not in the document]
ceyhunsen: Fixed in https://github.com/chainwayxyz/clementine/pull/381/commits/8e01f06f2679bdf0c1fecefb64b55f1b070a10b2.

core/src/actor.rs
atacann: I think its good, we can even check if scripts length is equal to txin length, because you might get the wrong script if they are not equal.

core/src/builder/sighash.rs
ceyhunsen: Fixed in https://github.com/chainwayxyz/clementine/pull/381/commits/48982ea43dc3e63993bc37c0219f6a78df42be52.

core/src/database/common.rs
```diff
@@ -139,6 +148,48 @@ impl Database {
...
+            Some(tx) => query.fetch_all(&mut **tx).await,
+            None => query.fetch_all(&self.connection).await,
+        };
+
+        match result {
+            Ok(operators) => {
+                // Check for missing indices
+                let indices: Vec<i32> = operators.iter().map(|(idx, _, _, _)| *idx).collect();
+                let expected_indices: Vec<i32> = (0..indices.len() as i32).collect();
+
```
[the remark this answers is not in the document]
ekrembal: it is consecutive starting from 0 in the initial setup.
But later, if some operators exit the system and we add more operators, these idxs can be discrete. But i think that fix would still require a code change, and it should be simple i guess

core/src/operator.rs
```diff
@@ -709,16 +710,32 @@ impl Operator {
...
+                let path = WinternitzDerivationPath {
+                    message_length: 40,
+                    log_d: 4,
+                    tx_type: crate::actor::TxType::BitVM,
+                    index: Some(self.idx as u32),
+                    operator_idx: None,
+                    watchtower_idx: None,
+                    time_tx_idx: Some(time_tx),
+                    intermediate_step_idx: Some(intermediate_step),
+                };
```
atacann, line 735: What is the difference between index and operator_idx?

core/src/builder/transaction.rs
```diff
@@ -300,6 +314,425 @@ pub fn create_kickoff_utxo_tx(
...
+    operator_unlock_hash: &[u8; 20],
+    nofn_xonly_pk: XOnlyPublicKey,
+    operator_xonly_pk: XOnlyPublicKey,
+    network: bitcoin::Network,
+) -> TxHandler {
+    let tx_ins = create_tx_ins(vec![
+        OutPoint {
+            txid: wcp_txid,
+            vout: watchtower_idx as u32,
+        },
```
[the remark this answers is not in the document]
atacann: vout is still not fixed, should be 0

ekrembal (approved):

PR #383, opened 2025-01-02
refactor(servers): improve gRPC server startup reliability

ekrembal:
Replace arbitrary sleep delays with proper server readiness detection using oneshot channels. This change:

- Removes unreliable fixed delays
- Adds proper server readiness detection
- Enables future shutdown handling capabilities
- Makes server startup more deterministic

The implementation now uses tokio::sync::oneshot channels to signal when each gRPC server is bound and ready to accept connections.


ceyhunsen (approved):
Replace this AI with me

PR #389, opened 2025-01-03
Parallel deposits for aggregator

### LINKED ISSUE #386: Multithreading in aggregator new deposit
# Issue Description

Currently, without finalizing the first signature, the aggregator won't get any new nonces; this is not what we want. Rather, we want to get nonces when available, aggregate them, get partial signatures when available, aggregate them, and finalize them.

- [ ] Collect nonces when available, don't wait for the next
- [ ] Aggregate collected nonce
- [ ] Collect partial signatures when available, don't wait for the next
- [ ] Aggregate collected partial signatures
- [ ] Finalize

https://github.com/chainwayxyz/clementine/blob/dev/core/src/rpc/aggregator.rs#L321 this for loop needs to be optimized

ceyhunsen:
# Description

Adds parallelism to `new_deposit` call in aggregator.

## Linked Issues

- Closes #386

### DIFF
--- a/core/src/errors.rs
+++ b/core/src/errors.rs
@@ -91,9 +91,12 @@ pub enum BridgeError {
     /// JSON RPC call failed
     #[error("JsonRpcError: {0}")]
     JsonRpcError(#[from] jsonrpsee::core::client::Error),
-    /// RPC interface requires a parameter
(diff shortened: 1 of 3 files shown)


ceyhunsen:
Closed in favor of #391

PR #394, opened 2025-01-08
TxHandler change to include output scripts

atacann:
Currently when we create connected tx's we need to recreate taproot scripts. It is easy to create bugs if any utxo is changed, as other tx's that use the changed utxo also needs to be changed. If output scripts are also in TxHandler other tx's can use this data and not recreate the scripts.

I didn't change any current utxo value or script in this PR.


core/src/builder/transaction.rs
```diff
@@ -23,12 +23,18 @@ use secp256k1::XOnlyPublicKey;
 pub struct TxHandler {
     /// Transaction itself.
     pub tx: bitcoin::Transaction,
+    /// Txid of the transaction, saved here to not repeatedly calculate it.
+    pub txid: Txid,
```
ceyhunsen, line 27: Do we really need this? I feel like only a handful of callers needs this information and they can save this information for their own use. For those who are not using, this is extra calculation and time waste IMO.

core/src/builder/transaction.rs
```diff
@@ -56,14 +62,14 @@ pub const OPERATOR_CHALLENGE_AMOUNT: Amount = Amount::from_sat(200_000_000);
 /// 2. Operator's Time Connector: timelocked utxo for operator for the entire withdrawal time
 /// 3. Kickoff input utxo: this utxo will be the input for the kickoff tx
 /// 4. P2Anchor: Anchor output for CPFP
-pub fn create_time_tx(
+pub fn create_time_txhandler(
```
ceyhunsen, line 65: Can you also update the function comment to match this?

core/src/builder/transaction.rs
```diff
@@ -246,15 +283,18 @@ pub fn create_move_tx_handler(
...
+        prev_scripts: vec![deposit_script],
+        prev_taproot_spend_infos: vec![Some(deposit_taproot_spend_info)],
+        out_scripts: vec![vec![], vec![]],
+        out_taproot_spend_infos: vec![Some(musig2_spendinfo), None],
     }
 }
 
 /// Creates the kickoff_tx for the operator. It also returns the change utxo
-pub fn create_kickoff_utxo_tx(
+pub fn create_kickoff_utxo_txhandler(
```
ceyhunsen, line 297: Comment update also required

ceyhunsen (changes requested):

core/src/builder/transaction.rs
atacann: I added this because we needed to keep creating Txid constantly and pass it to functions. I like that it is in TxHandler because it is very common. While creating all deposit tx's, you don't need to create Txid only for the tx's at the end like reimburse/disprove tx's. Maybe a good solution is change txid to Option<Txid>, and create a get_txid() function in TxHandler that calculates and stores txid if it is None, and returns it.

ceyhunsen: Didn't really track how deep txid needs to be passed tbh. It should be OK then.

Txid calculation should not be as expensive that it requires a lazy calculation logic. This should also be OK.

ekrembal: Txid calculation is double hash i think it is expensive

ceyhunsen (approved):

PR #423, opened 2025-01-17
Implement Verifier's Deposit Finalize handler

### LINKED ISSUE #354: Implement Verifier Deposit Finalize
### Issue Description

Really similiar to #344 but instead of signing, it will verify aggreagted signatures
We should consider caching calculated sighashes in an array

### Tasks

Parts of the task:
- [x] Do the same as #344 but verify signatures
- [ ] partially sign move tx
- [ ] save final aggregated signatures to the db
- [ ] _Save operator's signatures_ (TODO for future)
- [ ] impelement signature getter to use when needed

### Deadline of the Issue

TBD

mmtftr:
# Description

Created a new deposit_signatures table with primary key of (outpt, op_idx)
Implemented deposit finalize handler:
- Verifies all incoming signatures,
- Saves to DB
- Signs Move_TX and returns partial_sig

## Linked Issues

- Closes #354

## Testing

Currently no tests since I don't have a complete understanding of the business logic here.

> ![WARNING]
> Need to add tests before merge

### DIFF
--- a/core/src/database/common.rs
+++ b/core/src/database/common.rs
@@ -279,10 +279,10 @@ impl Database {
             "UPDATE deposit_kickoff_generator_txs
                 SET cur_unused_kickoff_index = cur_unused_kickoff_index + 1
                 WHERE id = (
-                    SELECT id 
-                    FROM deposit_kickoff_generator_txs 
+                    SELECT id
+                    FROM deposit_kickoff_generator_txs
                     WHERE cur_unused_kickoff_index < num_kickoffs
-                    ORDER BY id DESC 
+                    ORDER BY id DESC
                     LIMIT 1
                 )
                 RETURNING txid, raw_signed_tx, cur_unused_kickoff_index;", // This query returns the updated cur_unused_kickoff_index.
@@ -606,8 +606,8 @@ impl Database {
             .push_bind(OutPointDB(deposit_outpoint))
             .push(
                 " RETURNING nonces.internal_idx, sec_nonce, agg_nonce)
-            SELECT updated.sec_nonce, updated.agg_nonce 
-            FROM updated 
+            SELECT updated.sec_nonce, updated.agg_nonce
+            FROM updated
             ORDER BY updated.internal_idx;",
             )
             .build_query_as();
@@ -1141,6 +1141,56 @@ impl Database {
         }?;
 
         Ok(XOnlyPublicKey::from_slice(&xonly_key.0)?)
+    }
+
+    /// Saves the verified deposit signatures to the database
+    #[tracing::instrument(skip(self, tx, signatures), err(level = tracing::Level::ERROR), ret(level = tracing::Level::TRACE))]
+    pub async fn save_deposit_signatures(
+        &self,
+        tx: Option<&mut sqlx::Transaction<'_, Postgres>>,
+        deposit_outpoint: OutPoint,
+        operator_idx: u32,
+        signatures: Vec<schnorr::Signature>,
+    ) -> Result<(), BridgeError> {
+        let query = sqlx::query(
+            "INSERT INTO deposit_signatures (deposit_outpoint, operator_idx, signatures) VALUES ($1, $2, $3);"
+        )
+        .bind(OutPointDB(deposit_outpoint))
+        .bind(operator_idx as i64)
+        .bind(SignaturesDB(signatures));
+
+        match tx {
+            Some(tx) => query.execute(&mut **tx).await?,
+            None => query.execute(&self.connection).await?,
+        };
+
+        Ok(())
+    }
+
+    /// Retrieves the deposit signatures from the database
+    #[tracing::instrument(skip(self, tx), err(level = tracing::Level::ERROR), ret(level = tracing::Level::TRACE))]
+    pub async fn get_deposit_signatures(
+        &self,
+        tx: Option<&mut sqlx::Transaction<'_, Postgres>>,
+        deposit_outpoint: OutPoint,
+        operator_idx: u32,
+    ) -> Result<Option<Vec<schnorr::Signature>>, BridgeError> {
+        let query = sqlx::query_as(
+            "SELECT signatures FROM deposit_signatures WHERE deposit_outpoint = $1 AND operator_idx = $2;"
+        )
+        .bind(OutPointDB(deposit_outpoint))
+        .bind(operator_idx as i64);
+
+        let result: Result<(SignaturesDB,), sqlx::Error> = match tx {
+            Some(tx) => query.fetch_one(&mut **tx).await,
+            None => query.fetch_one(&self.connection).await,
+        };
+
+        match result {
+            Ok((SignaturesDB(signatures),)) => Ok(Some(signatures)),
+            Err(sqlx::Error::RowNotFound) => Ok(None),
+            Err(e) => Err(BridgeError::DatabaseError(e)),
+        }
     }
 }
 
--- a/core/src/rpc/verifier.rs
+++ b/core/src/rpc/verifier.rs
@@ -4,9 +4,11 @@ use super::clementine::{
     VerifierDepositSignParams, VerifierParams, VerifierPublicKeys, WatchtowerParams,
 };
 use crate::{
+    actor::Actor,
     builder::{
         self,
         sighash::{calculate_num_required_sigs, create_nofn_sighash_stream},
+        transaction::create_move_txhandler,
     },
     errors::BridgeError,
     musig2::{self, MuSigPubNonce, MuSigSecNonce},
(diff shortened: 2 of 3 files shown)


core/src/rpc/verifier.rs
```diff
@@ -531,10 +524,29 @@ impl ClementineVerifier for Verifier {
...
+            .save_deposit_signatures(None, deposit_outpoint, 0, verified_sigs)
+            .await?;
+
+        // Generate partial signature for move transaction
+        let mut move_txhandler = create_move_txhandler(
+            deposit_outpoint,
+            evm_address,
+            &recovery_taproot_address,
+            self.nofn_xonly_pk,
+            u32::try_from(user_takes_after).expect("User takes after is too large"),
```
ekrembal, line 538: Not sure if this is needed

core/src/rpc/verifier.rs
```diff
@@ -531,10 +524,29 @@ impl ClementineVerifier for Verifier {
...
+            self.config.bridge_amount_sats,
+            self.config.network,
+        );
 
+        let move_tx_sighash = Actor::convert_tx_to_sighash_script_spend(&mut move_txhandler, 0, 0)?;
+
+        let partial_sig = self.signer.sign(move_tx_sighash);
```
ekrembal, line 545: This should be a partial sign, so the musig nonce_gen should generate one more nonce and that nonce should be used in signing move tx

core/src/rpc/verifier.rs
```diff
@@ -531,10 +524,29 @@ impl ClementineVerifier for Verifier {
             }
         }
 
-        tracing::info!("Deposit finalized");
+        // Save signatures to database
+        self.db
```
ekrembal, line 528: We should split all the signatures to per operator list and save them
so verified_sigs.len() = k*operators.len()

ekrembal (changes requested):

core/src/rpc/verifier.rs
mmtftr: I opted to use fallible conversion here to avoid thinking about the overflow case.

This function taking u64 is likely incorrect (API uses u32 everywhere else), so I'll try to fix that and remove this conversion altogether. If I can't fix, I think it's fine to just truncate.

In general, I think it's important to consider the behavior of integer conversion especially to avoid bugs that might come up due to incorrect casting (on overflows/underflows etc)

core/src/rpc/verifier.rs
mmtftr: I can't partially sign without an agg_nonce here, how can I orchestrate that?

core/src/rpc/verifier.rs
mmtftr: Done

core/src/rpc/verifier.rs
mmtftr: Done

mmtftr:
There are some issues with the backend causing tests to fail. Other than that, I think it's ready for review.

ozankaymak:
Please resolve the conflicts before review.

mmtftr (commented):
Questions to the open

mmtftr:
> Please resolve the conflicts before review.

Conflicts have been resolved, waiting for review

core/src/rpc/clementine.proto
```diff
@@ -142,12 +144,11 @@ message NonceGenRequest {
 }
 
 message NonceGenFirstResponse {
+  reserved 3; // 3 = sig, previously used for signing new keypairs
   // Nonce ID
   uint32 id = 1;
   // New public key for the deposit
   bytes public_key = 2;
```
ekrembal, line 151: We don't need public key here

core/src/rpc/clementine.proto
```diff
@@ -142,12 +144,11 @@ message NonceGenRequest {
 }
 
 message NonceGenFirstResponse {
+  reserved 3; // 3 = sig, previously used for signing new keypairs
```
ekrembal, line 147: We don't need to be backward compatible now

core/src/rpc/verifier.rs
```diff
@@ -4,25 +4,64 @@ use super::clementine::{
...
+) -> Result<
+    (
+        bitcoin::OutPoint,
+        EVMAddress,
+        bitcoin::Address<NetworkUnchecked>,
+        u64,
+        u32,
+    ),
+    Status,
+> {
```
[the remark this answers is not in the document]
ekrembal: Sure, maybe we can create an issue for that to not forget?

core/src/rpc/verifier.rs
```diff
@@ -393,6 +398,7 @@ impl ClementineVerifier for Verifier {
         &self,
         req: Request<Streaming<VerifierDepositFinalizeParams>>,
     ) -> Result<Response<PartialSig>, Status> {
+        use clementine::verifier_deposit_finalize_params::Params;
```
ekrembal, line 401: Why `use` here?

ekrembal (approved):

core/src/rpc/aggregator.rs
```diff
@@ -481,11 +540,15 @@ impl ClementineAggregator for Aggregator {
...
+
+        // Start the nonce distribution pipe.
         let nonce_dist_handle = tokio::spawn(nonce_distributor(
             agg_nonce_receiver,
             partial_sig_streams,
             partial_sig_sender,
         ));
+
+        // Start the signature aggregation pipe.
         let sig_agg_handle = thread::spawn(|| {
```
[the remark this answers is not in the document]
ceyhunsen: It's because aggregation tasks are CPU dependent while distributor tasks are IO dependent. IO dependent tasks are opted for a Tokio "thread" to reduce memory usage and improve startup time. Tried using Tokio's `spawn` and `spawn_blocking`s but they failed to run 2 aggregation tasks in parallel, as far as I remember.

Also, we need all these tasks to run at the "same time" to achieve pipeline. That is the reason we spawn other Tokio runtimes inside this async function.

core/src/rpc/verifier.rs
mmtftr: easier to read and parse for the eyes below. These params are only relevant for this function, so we can put them in scope to make it easier to read below.

core/src/rpc/aggregator.rs
```diff
@@ -495,18 +558,25 @@ impl ClementineAggregator for Aggregator {
...
             final_sig_receiver,
             deposit_finalize_sender,
+            movetx_agg_nonce,
         ));
 
-        nonce_agg_handle.join().unwrap().unwrap();
+        // Wait for all pipeline tasks to complete
         try_join_all(vec![nonce_dist_handle]).await.unwrap();
         sig_agg_handle.join().unwrap().unwrap();
         try_join_all(vec![sig_dist_handle]).await.unwrap();
```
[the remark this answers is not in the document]
ceyhunsen: Yes, it looks like `try_join_all` calls are unnecessary. Can be replaced with smthng like `nonce_dist_handle.await.unwrap().unwrap()`.

core/src/rpc/aggregator.rs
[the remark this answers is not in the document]
mmtftr: I'm not sure if that's related, there really shouldn't be such a big difference unless there is a bug in Tokio itself. We can write a test for this to make sure it works properly

On a related note, I saw some problems with the mutex lock being held for a long time (i.e. for most of the duration of the function) in some functions. I believe I fixed one of these in the PR but we may benefit from testing whether locks are being held for too long to the point they'd reduce parallelism.

In this function's case, holding the sessions lock would block any other aggregator from running for any other session.

core/src/rpc/verifier.rs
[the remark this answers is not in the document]
mmtftr: #441 done

core/src/rpc/clementine.proto
mmtftr: removed

core/src/rpc/clementine.proto
mmtftr: removed

core/src/database/common.rs
```diff
@@ -1142,6 +1142,56 @@ impl Database {
 
         Ok(XOnlyPublicKey::from_slice(&xonly_key.0)?)
     }
+
+    /// Saves the verified deposit signatures to the database
+    #[tracing::instrument(skip(self, tx, signatures), err(level = tracing::Level::ERROR), ret(level = tracing::Level::TRACE))]
+    pub async fn save_deposit_signatures(
```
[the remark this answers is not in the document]
ozankaymak: Please add what we are saving here, similar for what we are retrieving.

core/src/rpc/clementine.proto
```diff
@@ -142,14 +144,10 @@ message NonceGenRequest {
 }
 
 message NonceGenFirstResponse {
```
ozankaymak, line 146: Why did we get rid of public_key?

mmtftr: all signatures will be done through verifier's keypair

core/src/database/common.rs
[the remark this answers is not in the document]
mmtftr: Done

PR #436, opened 2025-01-20
Txin sequence numbers

### LINKED ISSUE #422: Fix txin sequence numbers
### Issue Description

Right now every tx in transaction.rs is generated with inputs sequence field set as bitcoin::transaction::Sequence::ENABLE_RBF_NO_LOCKTIME. For every utxo with relative timelocks this needs to be changed

### Tasks

Parts of the task:
- [ ] Change txin sequence values in builder/transaction.rs

### Deadline of the Issue

.

atacann:
# Description

Use correct sequence numbers for Txin's with timelocks, also update timelock durations to current design 

## Linked Issues

- Closes #422

### DIFF
--- a/core/src/builder/transaction.rs
+++ b/core/src/builder/transaction.rs
@@ -65,13 +65,14 @@ pub fn create_sequential_collateral_txhandler(
     num_kickoffs_per_timetx: usize,
     network: bitcoin::Network,
 ) -> TxHandler {
-    let tx_ins = create_tx_ins(vec![OutPoint {
-        txid: input_txid,
-        vout: 0,
-    }]);
-    if max_withdrawal_time_block_count > u16::MAX as i64 {
-        panic!("Max withdrawal time block count is larger than u16::MAX (sequence relative timelock is 16 bits");
-    }
+    let tx_ins = create_tx_ins(
+        vec![OutPoint {
+            txid: input_txid,
+            vout: 0,
+        }]
+        .into(),
+    );
+    u16::try_from(max_withdrawal_time_block_count).expect("Max withdrawal time block count is larger than u16::MAX (sequence relative timelock is 16 bits");
     let max_withdrawal_time_locked_script = builder::script::generate_relative_timelock_script(
         operator_xonly_pk,
         max_withdrawal_time_block_count,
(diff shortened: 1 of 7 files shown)


ozankaymak (approved):

core/src/builder/transaction.rs
```diff
@@ -69,7 +69,9 @@ pub fn create_sequential_collateral_txhandler(
         txid: input_txid,
         vout: 0,
     }]);
-
+    if max_withdrawal_time_block_count > u16::MAX as i64 {
+        panic!("Max withdrawal time block count is larger than u16::MAX (sequence relative timelock is 16 bits");
```
ekrembal, line 73: is this needed?

ekrembal (changes requested):
Let's also move these 2.5 week, 2 week constants to some constants file. We can do this in a seperate pr. I will create an issue

core/src/builder/transaction.rs
```diff
@@ -1325,19 +1354,27 @@ pub fn create_tx_ins(outpoints: Vec<OutPoint>) -> Vec<TxIn> {
...
 }
 
-pub fn create_tx_ins_with_sequence(utxos: Vec<OutPoint>, height: u16) -> Vec<TxIn> {
-    let mut tx_ins = Vec::new();
-
-    for utxo in utxos {
-        tx_ins.push(TxIn {
+/// Create tx_ins with relative time locks using block height
+/// This function needs to be used with any TX that has a relative time lock in at least one of the inputs.
+pub fn create_tx_ins_with_sequence(utxos: Vec<OutPoint>, heights: &[Option<u16>]) -> Vec<TxIn> {
```
mmtftr, line 1359: In order of preference:

1. Newtype paradigm for better API, more rigid enforcement of contract:
```rust
pub struct TxInArgs {
  outpoint: OutPoint,
  height: Option<u16>,
}

impl From<OutPoint> for TxInArgs {
...
}

impl From<Vec<OutPoint>> for Vec<TxInArgs> {
...
}

impl From<Vec<(OutPoint, u16)>> for Vec<TxInArgs> {
...
}

// Single API with easy usage
// create_tx_ins(my_outpoint_vec.into())
// create_tx_ins(&my_tx_in_args_vec)
pub fn create_tx_ins(tx_in_args: impl AsRef<[TxInArgs]>) -> Vec<TxIn> {...}

// If you want to go one level higher:
// create_tx_ins(&my_outpoint_slice)
// Might cause too much codegen, so consider monomorphizing if going this route
pub fn create_tx_ins(tx_in_args: impl AsRef<[impl Into<TxInArgs>]>) -> Vec<TxIn> {...}

```

2. Use tuples to enforce that heights and utxos must have the same length:
```rust
pub fn create_tx_ins_with_sequence(tx_in_args: impl AsRef<[(OutPoint, Option<u16>)]>) -> Vec<TxIn> {...}
```

mmtftr (changes requested):

core/src/builder/transaction.rs
mmtftr: I'd prefer a constants module for things that'd change internal sighashes, #signatures. (basically anything that can cause errors in the deposit mechanism should not be configurable by the user)

Also, as @atacann said, we can create a test constants module that would provide very small defaults for fast tests.

mmtftr: If we're keeping this line, I think the following would be more readable (in fact, I think the behavior in your code may be incorrect as it doesn't check the lower bound):
```rust
u16::try_from(max_withdrawal_time_block_count).expect("Max withdrawal ...");
```

core/src/builder/transaction.rs
mmtftr: Feel free to skip this, as it's likely out of scope for this PR:

I'd prefer new module for utils like this, `builder::transaction::utils`. 1500 line file is not the best DX

ekrembal (approved):

PR #438, opened 2025-01-20
Save bitvm setup

### LINKED ISSUE #390: Save BitVM Setups on Bridge setup
### Issue Description

BitVM Setup takes minutes. We should store assert Tx taproot addresses and disprove tapscripthashes in the db to access easily when a deposit happens.

### Tasks

- [ ] schema for storing bitvm assert addresses and disprove script hashes
- [ ] update assert tx builders to use precalculated values

ekrembal:
Closes #390

### DIFF
--- a/core/src/actor.rs
+++ b/core/src/actor.rs
@@ -26,7 +26,7 @@ pub enum TxType {
 
 /// Derivation path specification for Winternitz one time public key generation.
 #[derive(Debug, Clone, Copy)]
-pub struct WinternitzDerivationPath {
+pub struct WinternitzDerivationPath<'a> {
     pub message_length: u32,
     pub log_d: u32,
     pub tx_type: TxType,
@@ -35,9 +35,9 @@ pub struct WinternitzDerivationPath {
     pub watchtower_idx: Option<u32>,
     pub time_tx_idx: Option<u32>,
     pub kickoff_idx: Option<u32>,
-    pub intermediate_step_idx: Option<u32>,
+    pub intermediate_step_name: Option<&'a str>,
 }
-impl WinternitzDerivationPath {
+impl<'a> WinternitzDerivationPath<'a> {
     fn to_vec(self) -> Vec<u8> {
         let index = match self.index {
             None => 0,
@@ -59,9 +59,9 @@ impl WinternitzDerivationPath {
             None => 0,
             Some(i) => i + 1,
         };
-        let intermediate_step_idx = match self.intermediate_step_idx {
-            None => 0,
-            Some(i) => i + 1,
+        let intermediate_step_name = match self.intermediate_step_name {
+            None => vec![],
+            Some(name) => name.as_bytes().to_vec(),
         };
 
         [
@@ -72,14 +72,14 @@ impl WinternitzDerivationPath {
                 watchtower_idx.to_be_bytes(),
                 time_tx_idx.to_be_bytes(),
                 kickoff_idx.to_be_bytes(),
-                intermediate_step_idx.to_be_bytes(),
             ]
             .concat(),
+            intermediate_step_name,
         ]
         .concat()
     }
 }
-impl Default for WinternitzDerivationPath {
+impl Default for WinternitzDerivationPath<'_> {
     fn default() -> Self {
         Self {
             message_length: Default::default(),
@@ -90,7 +90,7 @@ impl Default for WinternitzDerivationPath {
             watchtower_idx: Default::default(),
             time_tx_idx: Default::default(),
             kickoff_idx: Default::default(),
-            intermediate_step_idx: Default::default(),
+            intermediate_step_name: Default::default(),
         }
     }
 }
@@ -314,7 +314,7 @@ impl Actor {
     /// Returns derivied Winternitz secret key from given path.
     fn get_derived_winternitz_sk(
         &self,
-        path: WinternitzDerivationPath,
+        path: WinternitzDerivationPath<'_>,
     ) -> Result<winternitz::SecretKey, BridgeError> {
         let wsk = self
             .winternitz_secret_key
@@ -325,7 +325,7 @@ impl Actor {
     /// Generates a Winternitz public key for the given path.
     pub fn derive_winternitz_pk(
         &self,
-        path: WinternitzDerivationPath,
+        path: WinternitzDerivationPath<'_>,
     ) -> Result<winternitz::PublicKey, BridgeError> {
         let winternitz_params = winternitz::Parameters::new(path.message_length, path.log_d);
 
@@ -338,7 +338,7 @@ impl Actor {
     /// Signs given data with Winternitz signature.
     pub fn sign_winternitz_signature(
         &self,
-        path: WinternitzDerivationPath,
+        path: WinternitzDerivationPath<'_>,
         data: Vec<u8>,
     ) -> Result<Witness, BridgeError> {
         let winternitz = Winternitz::<BinarysearchVerifier, StraightforwardConverter>::new();
@@ -597,7 +597,7 @@ mod tests {
             .concat()
         );
 
-        params.intermediate_step_idx = Some(5);
+        params.intermediate_step_name = Some("step5");
         assert_eq!(
             params.to_vec(),
             [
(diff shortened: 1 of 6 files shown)


core/src/database/common.rs
```diff
@@ -1138,6 +1138,97 @@ impl Database {
...
+    #[tracing::instrument(skip(self), err(level = tracing::Level::ERROR), ret(level = tracing::Level::TRACE))]
+    pub async fn save_bitvm_setup(
+        &self,
+        tx: Option<&mut sqlx::Transaction<'_, Postgres>>,
+        operator_idx: i32,
+        time_tx_idx: i32,
+        kickoff_idx: i32,
+        assert_tx_addrs: Vec<Vec<u8>>,
+        root_hash: &[u8; 32],
+        public_input_wots: Vec<[u8; 20]>,
```
mmtftr, line 1152: consider impl AsRef<[u8; 20]>

also impl AsRef<[impl AsRef<[u8]>]> (not sure if this works)

core/src/actor.rs
```diff
@@ -35,9 +35,9 @@ pub struct WinternitzDerivationPath {
     pub watchtower_idx: Option<u32>,
     pub time_tx_idx: Option<u32>,
     pub kickoff_idx: Option<u32>,
-    pub intermediate_step_idx: Option<u32>,
+    pub intermediate_step_name: Option<&'a str>,
```
mmtftr, line 38: Does this have to be a &str? Can it be an enum?

core/src/rpc/verifier.rs
```diff
@@ -117,10 +135,102 @@ impl ClementineVerifier for Verifier {
...
+                    let (assert_tx_addr, _) = builder::address::create_taproot_address(
+                        &[script.clone()],
+                        None,
+                        self.config.network,
+                    );
+                    assert_tx_addr.script_pubkey().to_bytes()
+                })
+                .collect::<Vec<_>>();
+
+            // // This part is not working.
```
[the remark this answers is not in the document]
ekrembal: We should still use a correct verification key, rn we

core/src/builder/transaction.rs
```diff
@@ -615,37 +616,24 @@ pub fn create_operator_challenge_nack_txhandler(
...
-            builder::address::create_taproot_address(&[intermediate_script.clone()], None, network);
-        scripts.push(vec![intermediate_script]);
-        spendinfos.push(Some(intermediate_spend));
+
+    let mut txouts = vec![];
+    for i in 0..PARALLEL_ASSERT_TX_CHAIN_SIZE {
         txouts.push(TxOut {
-            value: Amount::from_sat(660), // TOOD: Hand calculate this
-            script_pubkey: intermediate_addr.script_pubkey(), // TODO: Add winternitz checks here
+            value: Amount::from_sat(330), // Minimum amount for a taproot output
```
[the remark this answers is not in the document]
ekrembal: Used MIN_TAPROOT_AMOUND

core/src/actor.rs
ekrembal: it is bcs BitVM uses strings to differentiate between intermediate values

core/src/builder/transaction.rs
```diff
@@ -663,84 +651,80 @@ pub fn create_assert_begin_txhandler(
...
+        last_mini_assert_txid[i] = mini_assert_tx.compute_txid();
+    }
 
-    let mut txins = (0..NUM_INTERMEDIATE_STEPS)
-        .map(|i| OutPoint {
-            txid: mini_assert_txhandlers[i].txid,
+    let mut txins = vec![];
+    for i in 0..PARALLEL_ASSERT_TX_CHAIN_SIZE {
+        txins.push(OutPoint {
+            txid: last_mini_assert_txid[i],
```
atacann, line 706: txids here should be last mini asserts

core/src/builder/transaction.rs
```diff
@@ -764,17 +748,20 @@ pub fn create_assert_end_txhandler(
     let assert_end_tx = create_btc_tx(create_tx_ins(txins), tx_outs);
 
     let mut prevouts = (0..NUM_INTERMEDIATE_STEPS)
```
atacann, line 750: should be PARALLEL_ASSERT_TX_CHAIN_SIZE instead of NUM_INTERMEDIATE_STEPS, for the following loops too

ekrembal: Done

core/src/builder/transaction.rs
ekrembal: What do you mean?

atacann: Ah, I think this is right if you want to hold last PARALLEL_ASSERT_TX_CHAIN_SIZE txs in last_mini_assert_txid.
But then you have to use modulo in the loop above
``
create_mini_assert_tx(
            last_mini_assert_txid[i % PARALLEL_ASSERT_TX_CHAIN_SIZE],
``
and 
`last_mini_assert_txid[i % PARALLEL_ASSERT_TX_CHAIN_SIZE] = mini_assert_tx.compute_txid();`

ekrembal: I do it with if/else. Is there a bug?

core/src/builder/address.rs
```diff
@@ -7,13 +7,32 @@ use crate::builder;
...
 use bitcoin::address::NetworkUnchecked;
+use bitcoin::hashes::Hash;
 use bitcoin::Amount;
 use bitcoin::{
     secp256k1::XOnlyPublicKey,
     taproot::{TaprootBuilder, TaprootSpendInfo},
     Address, ScriptBuf,
 };
 
+pub fn create_taproot_builder(scripts: &[ScriptBuf]) -> TaprootBuilder {
```
mmtftr, line 18: would prefer to have some docs here: what kind of tree does it construct (balanced or not etc)

Also impl AsRef<>

I would also rename to something like `taproot_builder_with_scripts` as it's not just creating a taproot builder but including scripts as well

ceyhunsen (changes requested):

core/src/builder/address.rs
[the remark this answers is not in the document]
ekrembal: This would not give any errors, so no need for Result

core/src/database/common.rs
```diff
@@ -1138,6 +1138,98 @@ impl Database {
...
+
+    /// Saves BitVM setup data for a specific operator, time_tx and kickoff index combination
+    // #[tracing::instrument(skip(self), err(level = tracing::Level::ERROR), ret(level = tracing::Level::TRACE))]
+    pub async fn save_bitvm_setup(
+        &self,
+        tx: Option<&mut sqlx::Transaction<'_, Postgres>>,
+        operator_idx: i32,
+        time_tx_idx: i32,
+        kickoff_idx: i32,
+        assert_tx_addrs: impl AsRef<[Vec<u8>]>,
```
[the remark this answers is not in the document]
ekrembal: Let's do AsRef<[ScriptBuf]> i just want to store scriptpubkey of assert txs

core/src/database/common.rs
```diff
@@ -1836,4 +1928,49 @@ mod tests {
...
+        assert_eq!(result.0, assert_tx_addrs);
+        assert_eq!(result.1, root_hash);
+        assert_eq!(result.2, public_input_wots);
+
+        // Test non-existent entry
+        let non_existent = database
+            .get_bitvm_setup(None, 999, time_tx_idx, kickoff_idx)
+            .await
+            .unwrap();
+        assert!(non_existent.is_none());
```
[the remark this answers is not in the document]
ekrembal: In db functions, we return None if there is no records found. BitvmSetupNotFound is a different error

core/src/builder/address.rs
```diff
@@ -7,13 +7,32 @@ use crate::builder;
...
+    let n = scripts.len();
+    if n == 0 {
+        TaprootBuilder::new()
```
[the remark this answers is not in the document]
ekrembal: Maybe in future prs. I didn't add any new codes, just moved some to another function to be able to use it.

core/src/builder/address.rs
ekrembal: Done

core/src/builder/transaction.rs
ekrembal: Done, good catch!

core/src/database/common.rs
ekrembal: Done!

Issue #431, opened 2025-01-20
Write documentation


ozankaymak:
### Proposal Description

For both easier onboarding of new people and better DX, we need to write more documentation.

mmtftr:
I believe we can approach this systematically. Using Rustdoc is the industry standard for documenting a crate, so I believe we can start from there. We can use this as a tracker issue to track work on documenting the codebase.

A couple of candidate tasks for the first stage:
- Auto-generated and hosted rustdoc ([this looks like a good starting point](https://github.com/KGrewal1/gauss/blob/master/.github/workflows/rust-doc.yml))
- README improvements like:
  - including an overview of the bridge, linking to important documents and concepts (BitVM bridge, MuSig2, P2TR, classical [transactions](https://developer.bitcoin.org/devguide/transactions.html), [bitcoin contracts](https://developer.bitcoin.org/devguide/contracts.html) etc.)

mmtftr:
- Another great thing would be to have a CONTRIBUTING.md that both introduces the project on an overview-level and provides guidance on how to contribute (our testing, documentation, coverage standards, PR contents, commit formatting, preferred PR merge, branch naming convention, branch explanations [dev vs main etc] etc.).
- On a similar note, it'd be great if we adopt [Doctesting](https://doc.rust-lang.org/rust-by-example/testing/doc_testing.html), especially for simpler functions where the documentation could provide sufficient tests. A lot of unit tests could probably be rolled into doctests. Of course, some functions are way too complicated to test within doctests.

ceyhunsen:
- I think it will be more appropriate to keep concepts in other documentations, outside of Clementine, like Clementine onboarding docs
  - Or in a separate dir called `docs/`
  - Readme can just include small intro to Clementine and how to run it
- #364 contributing.md
- I believe we can slowly shift basic unit tests to doctests and add rustdoc support
  - It should be very good if we adopt it but there is no rush to do that rn
  - Everyone slowly switching to this will help having this habit for everyone of us, rather than single person doing all the work
  - I suggest we should push ourselves to do this in our PR's for the new codes
  - It will be nice to move basic unit tests to doctest, if one's PR touches that specific part of code or in the same context

Issue #410, opened 2025-01-13
Clean code


ekrembal:
### Issue Description

We should clean the code for better dev-experience

### Tasks

Get rid of:
- [x] hashes.rs
- [x] merkle.rs
- [ ] user.rs (User's wont be using this, this is just for testing)
- [x] unused functions in utils.rs
- [x] remove sighash extractor functions in actor.rs and move them to sighash builder
- [x] env_writer.rs

### Deadline of the Issue

-

ozankaymak:
Closed in favor of #454

PR #439, opened 2025-01-20
Clean code, better documentation and comments

ozankaymak:
# Description
This PR introduces better documentation and code comments. Also cleans the code.
## Linked Issues
#410 
#431 

## Testing
None

## Docs
This whole PR is about the docs.


core/src/builder/sighash.rs
```diff
@@ -19,6 +26,7 @@ pub fn calculate_num_required_sigs(config: &BridgeConfig) -> usize {
         * (10 + 2 * config.num_watchtowers)
 }
 
+/// Generates the sighash for a given transaction input for key spend path.
 pub fn convert_tx_to_pubkey_spend(
```
mmtftr, line 30: this is bad naming, it's not converting a tx but it's generating a sighash with a pubkey spend for this transaction's i'th input.

Since it basically converts our API types in a format suitable to run `taproot_key_spend_signature_hash`, we can call it `calculate_key_spend_signature_hash`, also it'd be even better to define it in the impl of TxHandler:

```rs
impl TxHandler {
  pub fn convert_tx_to_pubkey_spend(&mut self, txin_index: usize, sighash_type: Option<TapSighashType>) -> ...
}
```

mmtftr: ditto for all the ones below

mmtftr: > Since it basically converts our API types in a format suitable to run `taproot_key_spend_signature_hash` ...

I think it'd be even better to say that it does so in the rustdoc, and link to bitcoin's rustdoc, which I believe can be done like
```rs
/// [`bitcoin::sighash::SighashCache::taproot_key_spend_signature_hash`] for more details
```

core/src/builder/sighash.rs
```diff
@@ -158,45 +179,50 @@ pub fn create_nofn_sighash_stream(
...
 
+                    // Creates the start_happy_reimburse_tx handler.
                     let mut start_happy_reimburse_txhandler = builder::transaction::create_start_happy_reimburse_txhandler(
                         &kickoff_txhandler,
                         *operator_xonly_pk,
                         network
                     );
 
-                    // sign kickoff_tx utxo
+                    // Creates the sighash for the start_happy_reimburse_tx.
```
mmtftr, line 202: nit: yield sighash instead of create

core/src/builder/sighash.rs
```diff
@@ -80,9 +90,14 @@ pub fn convert_tx_to_script_spend(
     Ok(sig_hash)
 }
 
-/// Construct every deposit tx for each operator, sequential_collateral, and kickoff utxo,
-/// and yield the sighash for each txin that needs a NofN signature.
-/// Refer to bridge design diagram to see which NofN signatures are needed (the ones marked with blue arrows)
+/// For a given deposit tx, for each operator and sequential_collateral tx, generates the SigHash stream for:
```
mmtftr, line 93: A little more explanation for those that are unfamiliar:

...generates sighashes which are later used to sign with the NofN verifiers through the MuSig2 algorithm. Currently the following transactions are yielded by this stream (subject to change)

core/src/operator.rs
mmtftr, line 1: Why is everything commented out in this file? Can we not delete it?

mmtftr (changes requested):
Thanks a lot for undertaking this large task, I have some comments that might be helpful going forward.

core/src/builder/script.rs
```diff
@@ -63,23 +67,11 @@ pub fn create_deposit_script(
...
-
 /// ATTENTION: If you want to spend a UTXO using timelock script, the
-/// condition is that (`# in the script`) < (`# in the sequence of the tx`)
-/// < (`# of blocks mined after UTXO`) appears on the chain.
+/// condition is that (`# in the script`) ≤ (`# in the sequence of the tx`)
+/// ≤ (`# of blocks mined after UTXO appears on the chain`).
 pub fn generate_relative_timelock_script(
-    actor_taproot_xonly_pk: XOnlyPublicKey, // This is the tweaked XonlyPublicKey, which appears in the script_pubkey of the address
+    actor_taproot_xonly_pk: XOnlyPublicKey, // This is the tweaked XonlyPublicKey, which appears in the script_pubkey of the address. The tweaked signature will be given accordingly.
     block_count: i64,
```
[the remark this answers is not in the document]
ozankaymak: Can you explain this a little bit? I am confused.

core/src/builder/script.rs
```diff
@@ -63,23 +67,11 @@ pub fn create_deposit_script(
...
-        .push_x_only_key(&operator_xonly_pk)
-        .push_opcode(OP_CHECKSIG)
-        .into_script()
-}
-
 /// ATTENTION: If you want to spend a UTXO using timelock script, the
-/// condition is that (`# in the script`) < (`# in the sequence of the tx`)
-/// < (`# of blocks mined after UTXO`) appears on the chain.
+/// condition is that (`# in the script`) ≤ (`# in the sequence of the tx`)
+/// ≤ (`# of blocks mined after UTXO appears on the chain`).
```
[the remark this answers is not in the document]
ozankaymak: Done, also added that in fact using script locking is not mandatory.

core/src/operator.rs
ceyhunsen: It is probably because they are outdated (from old design and jsonrpc days). But they are valuable to leave as reference, soon to be updated to the new design.

core/src/builder/sighash.rs
ozankaymak: Done

core/src/builder/sighash.rs
ozankaymak: Done

core/src/operator.rs
mmtftr: I think it'd be better to leave a permalink to the original code instead: 
https://github.com/chainwayxyz/clementine/blob/c0c61e8f7e1f458e71a95ea5434401ea4e1185ad/core/src/operator.rs

Otherwise reviewing the rewrite PR might become more difficulty

core/src/builder/script.rs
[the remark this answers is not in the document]
mmtftr: The `nSequence: i64` can be used to represent many things (relative block count, relative time in seconds, disabling locktime, enabling RBF):
https://learnmeabitcoin.com/technical/transaction/input/sequence/

Here, the function expects a block_count, so it'd be better to force the i64 to be a block_count either by forcing it to be a u16 (the range where it represents a block count) or asserting that it is a block count within the function (with `u16::try_from(n_sequence)` where `n_sequence` is the input.

Otherwise, this function would allow arbitrary sequence values to be used, in which case the name should reflect that.

The most idiomatic way to do this would be a newtype (which is what bitcoin uses: https://docs.rs/bitcoin/latest/bitcoin/struct.Sequence.html)

core/src/builder/sighash.rs
ozankaymak: Done

core/src/builder/script.rs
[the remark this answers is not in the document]
ozankaymak: Done

core/src/builder/script.rs
```diff
@@ -63,34 +67,27 @@ pub fn create_deposit_script(
...
-}
-
 /// ATTENTION: If you want to spend a UTXO using timelock script, the
-/// condition is that (`# in the script`) < (`# in the sequence of the tx`)
-/// < (`# of blocks mined after UTXO`) appears on the chain.
+/// condition is that (`# in the script`) ≤ (`# in the sequence of the tx`)
+/// ≤ (`# of blocks mined after UTXO appears on the chain`). However, this is not mandatory.
+/// One can spend an output delayed for some number of blocks just by using the nSequence field
+/// of the input inside the transaction. For more, see:
+/// https://github.com/bitcoin/bips/blob/master/bip-0068.mediawiki and
```
ceyhunsen, line 75: Could be bullet list

ekrembal (approved):

core/src/builder/script.rs
```diff
@@ -63,34 +67,27 @@ pub fn create_deposit_script(
...
+/// https://github.com/bitcoin/bips/blob/master/bip-0112.mediawiki
 pub fn generate_relative_timelock_script(
-    actor_taproot_xonly_pk: XOnlyPublicKey, // This is the tweaked XonlyPublicKey, which appears in the script_pubkey of the address
-    block_count: i64,
+    actor_taproot_xonly_pk: XOnlyPublicKey, // This is the tweaked XonlyPublicKey, which appears in the script_pubkey of the address. The tweaked signature will be given accordingly.
```
ceyhunsen, line 78: i think this should be in the function comment so that everyone can see this comment without locating the code

core/src/operator.rs
ozankaymak: Good point, but we will leave them as they are for now.

core/src/builder/sighash.rs
```diff
@@ -159,113 +124,128 @@ pub fn create_nofn_sighash_stream(
...
 
-                    yield convert_tx_to_pubkey_spend(
-                        &mut challenge_tx,
+                    // Yields the sighash for the challenge_tx.input[0], which spends kickoff_tx.input[1] using SinglePlusAnyoneCanPay.
+                    yield challenge_tx.calculate_pubkey_spend_sighash(
                         0,
                         Some(bitcoin::sighash::TapSighashType::SinglePlusAnyoneCanPay)
                     )?;
 
+                    // Creates the start_happy_reimburse_tx handler.
```
ceyhunsen, line 139: a bit verbose imo

ozankaymak:
I will merge this PR as it is now, and distribute the remaining tasks accordingly.

ceyhunsen (approved):

core/src/constants.rs
```diff
@@ -19,3 +19,10 @@ pub type VerifierChallenge = (BlockHash, U256, u8);
 pub const NUM_INTERMEDIATE_STEPS: usize = 100;
 
 pub const PARALLEL_ASSERT_TX_CHAIN_SIZE: usize = 10;
```
mmtftr, line 21: short comment on what the constants are, eg:

/// Number of parallel assert txs to be executed after ...

core/src/builder/script.rs
ozankaymak: Done

core/src/builder/script.rs
ozankaymak: Done

core/src/constants.rs
ozankaymak: Done

core/src/builder/transaction.rs
```diff
@@ -254,28 +336,22 @@ pub fn create_move_tx(
...
         script_pubkey: musig2_address.script_pubkey(),
     };
 
-    // create_btc_tx(tx_ins, vec![move_txout, anyone_can_spend_txout])
     create_btc_tx(tx_ins, vec![move_txout, anchor_output])
 }
 
-/// Creates a [`TxHandler`] for the move_tx.
+/// Creates a [`TxHandler`] for the `move_to_vault_tx`.
 pub fn create_move_txhandler(
```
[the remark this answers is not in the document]
ozankaymak: I think create_move_to_vault_tx is needed since we only need the transaction and its id sometimes.

core/src/builder/transaction.rs
```diff
@@ -651,6 +664,8 @@ pub fn create_operator_challenge_nack_txhandler(
     }
 }
 
+/// Creates a [`TxHandler`] for the `operator_challenge_ACK_tx`. This transaction will allow the operator
+/// to send the `assert_begin_tx` to basically respond to the challenge(s).
```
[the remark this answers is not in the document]
ozankaymak: Done

core/src/builder/transaction.rs
```diff
@@ -1187,6 +1228,8 @@ pub fn create_reimburse_txhandler(
     }
 }
 
+/// Creates a [`TxHandler`] for the `kickoff_timeout_tx`. This transaction will be sent by the operator
+/// if there are no challenges, to be able to send `reimburse_tx` later.
```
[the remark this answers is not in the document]
ozankaymak: Done

atacann (approved):

Issue #432, opened 2025-01-20
Consider `impl AsRef` instead of Owned Vecs in API


mmtftr:
### Proposal Description

# Description

Our internal APIs often require passing in owned objects and clone them multiple times. Instead, to avoid copy costs, we should attempt to take `impl AsRef` or `impl Into<Cow>` (if an owned object is needed regardless), reducing the explicit clone requirements on the API caller whenever possible. An example is provided below from the (old) musig2 module:

```rust
/// Unmodified
#[tracing::instrument(ret(level = tracing::Level::TRACE))]
pub fn aggregate_nonces(pub_nonces: Vec<MuSigPubNonce>) -> MuSigAggNonce {
    let musig_pub_nonces: Vec<musig2::PubNonce> = pub_nonces
        .iter()
        .map(|x| musig2::PubNonce::from_bytes(&x.0).unwrap())
        .collect::<Vec<musig2::PubNonce>>();
    let musig_agg_nonce: AggNonce = AggNonce::sum(musig_pub_nonces);
    ByteArray66(musig_agg_nonce.into())
}

/// This one has **less** copies as it passes an iterator (without allocating a Vec). 
/// It also asks for only what is used, which is a ref to a slice of PubNonces. We'll end up
/// copy'ing due to `from_bytes` anyway, so the caller doesn't need to pass an owned
/// PubNonce. If we force the caller to pass an owned Vec, then the caller would need to
/// clone on their end as well.
pub fn aggregate_nonces_nocopy(pub_nonces: impl AsRef<[MuSigPubNonce]>) {
    let _musig_agg_nonce: AggNonce = AggNonce::sum(
        pub_nonces
            .as_ref()
            .iter()
            .map(|x| musig2::PubNonce::from_bytes(&x.0).unwrap()),
    );
}
```

Although this example might be less significant (as pubnonces are discarded quickly after aggregating), the same pattern is used in many interfaces. We should standardize our API to ask for only what is needed (ref in this case) and return as much as it can (in general)

ceyhunsen:
Fixed with #449

PR #449, opened 2025-01-21
Convert Owned Vec to Ref in API signatures

mmtftr:
# Description

Changed into &[] types for Vec taking functions:

- aggregate_partial_signatures
- create_assert_end_tx_handler
- aggregate_nonces

## Linked Issues

- Work on #432 (not complete)

Requires merge of #447 before review

### DIFF
--- a/core/src/builder/transaction.rs
+++ b/core/src/builder/transaction.rs
@@ -500,7 +500,7 @@ pub fn create_watchtower_challenge_kickoff_txhandler(
     kickoff_tx_handler: &TxHandler,
     num_watchtowers: u32,
     watchtower_xonly_pks: &[XOnlyPublicKey],
-    watchtower_wots: Vec<Vec<[u8; 20]>>,
+    watchtower_wots: &[Vec<[u8; 20]>],
     network: bitcoin::Network,
 ) -> TxHandler {
     let tx_ins = create_tx_ins(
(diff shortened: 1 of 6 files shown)


core/src/builder/transaction.rs
```diff
@@ -500,7 +500,7 @@ pub fn create_watchtower_challenge_kickoff_txhandler(
     kickoff_tx_handler: &TxHandler,
     num_watchtowers: u32,
     watchtower_xonly_pks: &[XOnlyPublicKey],
-    watchtower_wots: Vec<Vec<[u8; 20]>>,
+    watchtower_wots: &[Vec<[u8; 20]>],
```
ozankaymak, line 503: Out of curiosity, is it possible to do the same for the inner Vectors?

ozankaymak (approved):

core/src/builder/transaction.rs
mmtftr: It's more complicated. Taking a ref for the inner vector will cause more copies.

In some cases, we can `&[impl AsRef<[u8;20]>]` but when the downstream API doesn't support that we may have to clone anyway. I don't remember the exact reason here but I had encountered some problems in some of the APIs and left them as is for now. 

A lot of downstream APIs have inflexible API signatures that require cloning, so I tried as much as possible to mirror the downstream API that we have to use (to avoid unnecessary copies)

PR #453, opened 2025-01-21
style: minor cleanup

doscortados:
- address `clippy` warnings
- remove commented code (constants)


ceyhunsen:
Thanks for the interest. Because Clementine is in heavy development, so there are some code left commented out intentionally. Therefore we can't let you remove code without a valid reason.

PR #455, opened 2025-01-22
Streamed get_params for operators and watchtowers

### LINKED ISSUE #426: Implement gRPC stream for Operator and Watchtower get params
With the addition of multiple kickoff_txs per sequential_collateral_tx, we hit the 4mb limit of gRPC.
So let's convert them to gRPC streams

ceyhunsen:
- Closes #426


core/src/rpc/aggregator.rs
```diff
@@ -335,8 +341,14 @@ impl ClementineAggregator for Aggregator {
...
             async move {
+                let (tx, rx) = tokio::sync::mpsc::channel(1280);
+
+                client
+                    .set_operator(tokio_stream::wrappers::ReceiverStream::new(rx))
+                    .await?;
+
                 for param in params {
-                    client.set_operator(Request::new(param)).await?;
+                    tx.send(param).await.unwrap();
```
ceyhunsen, line 351: I found this part the most sketchy; Need a second view.

ekrembal (approved):
LGTM. This solves the problem of max message limit. But i think we can do this in a smarter by just doing the same thing as we did in new_deposit in aggregator. I will create an issue for this.

ceyhunsen:
Tests currently fail, don't merge.

core/src/rpc/verifier.rs
```diff
@@ -155,11 +169,27 @@ impl ClementineVerifier for Verifier {
...
+        for _ in 0..self.config.num_kickoffs_per_timetx
+            * self.config.num_time_txs
+            * utils::ALL_BITVM_INTERMEDIATE_VARIABLES.len()
+        {
+            let operator_params = in_stream
+                .message()
+                .await?
+                .ok_or(Status::invalid_argument("No first message received"))?
+                .response
+                .ok_or(Status::invalid_argument("No deposit outpoint received"))?;
```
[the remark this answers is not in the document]
ceyhunsen: Fixed in https://github.com/chainwayxyz/clementine/pull/455/commits/60d637c1bf146bb0b22d996e1633a77d4bd1d33e

core/src/rpc/verifier.rs
```diff
@@ -309,11 +365,23 @@ impl ClementineVerifier for Verifier {
...
 
-        let xonly_pk = XOnlyPublicKey::from_slice(&watchtower_params.xonly_pk).map_err(|_| {
+        let xonly_pk = in_stream
+            .message()
+            .await?
+            .ok_or(Status::invalid_argument("No message is received"))?
+            .response
+            .ok_or(Status::invalid_argument("No message is received"))?;
+        let xonly_pk = if let watchtower_params::Response::XonlyPk(xonly_pk) = xonly_pk {
+            xonly_pk
```
atacann, line 379: Maybe there is a better way to do these because we copy paste these a lot. Something with into maybe. Efe could tell us how 😄

atacann (approved):
I think it looks good but I am not familiar with grpc boilerplate. Maybe one suggestion would be to increase config.num_time_txs and/or config.num_kicksoffs_per_timetx in one test so that params exceed 4mb but I think if it works in smaller sizes it should work in larger ones too.

core/src/rpc/watchtower.rs
```diff
@@ -1,91 +1,57 @@
...
-        create_actors,
-        rpc::clementine::{clementine_watchtower_server::ClementineWatchtower, Empty},
-        watchtower::Watchtower,
-    };
-    use std::{env, thread};
-    use tonic::Request;
-
-    #[tokio::test]
-    #[serial_test::serial]
-    async fn watchtower_get_params() {
```
ozankaymak, line 57: Is it safe to get rid of this test?

ceyhunsen: Yes. Test was going to get a bit complex then it needed to be and this function is getting tested in different places.

I do think we need to refactor all this grpc stuff to be easily unit testable in the future. Then we can add this kinds of tests again.

PR #462, opened 2025-01-23
operator deposit

### LINKED ISSUE #397: Implement operator deposit_sign
### Issue Description

Operators sign 3 things per every time tx for a deposit.
These are Operator's burn connector input for Kickoff Timeout Tx and Disprove Tx, Already Disproved Tx.

Aggregator should collect these signatures form all operators, and when calling deposit_finalize on verifiers, aggregator also sends these operators' signatures as well.

### Tasks

Parts of the task:
- [ ] Create another sighash stream generator for sighashes that operator needs to sign
- [ ] Implement Kickoff Timeout Tx signature yield
- [ ] Implement Disprove tx signature yield
- [ ] Implement Already Disproved Tx signature yield
- [ ] Integrate Operator Deposit Sign to the Aggregator.
- [ ] Integarte operators' signatures verfying on verifiers deposit_finalize and saving to db

atacann:
# Description

Adds a sighash stream for operators that returns sighashs that needs to be signed by them.
Implement new_deposit() for operators that share their signatures
Aggregator collects and shares operator signatures to verifiers
Verifiers verify the signatures

Additionally fixes a bug in Database::get_operators() and creates a test for set and get_operators

## Linked Issues

- Closes #397


ekrembal (approved):
LGTM

core/src/database/common.rs
```diff
@@ -182,14 +182,17 @@ impl Database {
                 // Convert the result to the desired format
                 let data = operators
                     .into_iter()
-                    .map(|(_, pk, addr, txid)| {
-                        let xonly_pk = XOnlyPublicKey::from_str(&pk).unwrap();
-                        let addr = bitcoin::Address::from_str(&addr).unwrap().assume_checked();
-                        let txid = Txid::from_str(&txid).unwrap();
+                    .map(|(_, pk, addr, txid_db)| {
+                        let xonly_pk = XOnlyPublicKey::from_str(&pk).map_err(|e| {
```
ozankaymak, line 186: Why are we using Strings for XOnlyPublicKeys and Addresses? Regardless of this PR, we should change this, I think it makes the database operations slower.

core/src/database/common.rs
```diff
@@ -182,14 +182,17 @@ impl Database {
...
+                            BridgeError::Error(format!("Invalid XOnlyPublicKey: {}", e))
+                        })?;
+                        let addr = bitcoin::Address::from_str(&addr)
```
ozankaymak, line 189: Same here

core/src/rpc/aggregator.rs
```diff
@@ -282,6 +289,97 @@ impl Aggregator {
...
+            movetx_agg_nonce,
+            &musig_partial_sigs,
+            Message::from_digest(sighash.to_byte_array()),
+        )
+        .map_err(|x| BridgeError::Error(format!("Aggregating MoveTx signatures failed {}", x)))?;
+
+        // everything is fine, return the signed move tx
+        let _move_tx = move_txhandler.tx;
+        // TODO: Sign the transaction correctly after we create taproot witness generation functions
+        Ok(RawSignedMoveTx { raw_tx: vec![1, 2] })
```
ozankaymak, line 381: Why is it left like this?

core/src/rpc/aggregator.rs
```diff
@@ -429,35 +527,14 @@ impl ClementineAggregator for Aggregator {
 
         // Extract and validate deposit parameters
         let deposit_params = deposit_params_req.get_ref().clone();
```
ozankaymak, line 529: Please change the function documentation if needed.

ozankaymak (changes requested):
I think we need to change the NofNs from KeySpend to ScriptSpend if we want the tests to pass since we added the signature checks here.

core/src/database/common.rs
ekrembal: Not a high priority rn i think. We don't often push outpoint or address

core/src/rpc/aggregator.rs
ekrembal: We left tx generations to the next milestone

ekrembal: Or we can make it in a different pr. This Pr is about signatures that are collected from operators

Issue #454, opened 2025-01-22
Remove user.rs and find a better way to use them


ozankaymak:
### Issue Description

We will not be using user.rs anymore, but we still need their functionalities for testing.

### Tasks

Parts of the task:
- [ ] Remove user.rs
- [ ] Find a better way to include them in the tests

### Deadline of the Issue

-

ceyhunsen:
It should be noted that all the test code **must** be macros. Because we use them in both unit and integration tests and one of them must include the file directly (like by specifying relative OS path to it), which was the integration tests. Reason was that we don't want to infiltrate test code into binaries (details: https://github.com/chainwayxyz/clementine/issues/181).

If we use `use` statements, integration tests won't compile because include paths were different (crate:: vs clementine_core::). So we decided to use only macros for this kind of stuff. They are in `core/src/test_utils.rs`. I suggest following the conventions. Pr that introduced this: https://github.com/chainwayxyz/clementine/pull/375

ceyhunsen:
Closed by #472

PR #464, opened 2025-01-24
tests: use embedded postgres db

doscortados:
Use embedded postgres database for unit-testing using `pgtemp`.

Auxilary changes:
- make test setup macros into regular functions (simplifies imports & testing significantly)
- avoid usage of `serial_test` workaround (it was likely used to avoid race conditions I assume)
- update GitHub Action scripts to provide necessary postgres binaries and libraries


core/Cargo.toml
```diff
@@ -36,9 +36,9 @@ async-stream = { workspace = true }
 futures-util = { workspace = true }
 futures-core = { workspace = true }
 bitvm = { workspace = true }
+pgtemp = "0.5.0"
 
 [dev-dependencies]
-serial_test = { workspace = true }
```
ceyhunsen, line 41: We need serial testing for several reasons:

- We only use a single instance of Bitcoin regtest and some of the tests are fragile to changes outside of that test (e.g. some test mines a block while some other depends on it to not for a while)
- GRPC servers currently uses the same ports (not randomly assigned), therefore we can't spawn servers in parallel

But it is not related to database in any way: We start a clean db for every test.

core/src/lib.rs
```diff
@@ -27,8 +27,7 @@ pub mod utils;
 pub mod verifier;
 pub mod watchtower;
 
-#[cfg(test)]
```
ceyhunsen, line 30: This is dangerous! Please see

- https://github.com/chainwayxyz/clementine/issues/181
- https://github.com/chainwayxyz/clementine/pull/375
- https://github.com/chainwayxyz/clementine/issues/454#issuecomment-2607033070

ceyhunsen (changes requested):

core/src/actor.rs
```diff
@@ -574,7 +569,7 @@ mod tests {
 
     #[tokio::test]
     async fn derive_winternitz_pk_uniqueness() {
-        let config = create_test_config_with_thread_name!(None);
```
mmtftr, line 577: Please check [this comment](https://github.com/chainwayxyz/clementine/issues/454#issuecomment-2607033070) for why we can't use functions instead of macros.

mmtftr:
Are we going through with this PR @doscortados?

core/src/actor.rs
doscortados: I read the comment and still don't get why use so cumbersome and unfriendly way instead of just creating a sub-create "clementine-testkit" where all test utilites can be regular functions and still not be included in the main binary.

doscortados:
> Are we going through with this PR @doscortados?

Seems like too much now.

PR #472, opened 2025-01-27
Remove user

### LINKED ISSUE #454: Remove user.rs and find a better way to use them

ceyhunsen:
# Description

Moves test code from `User` to `test_utils` and removes `User.rs`. Additional notes are added to the lines that uses the old user module. These lines are not comments and will fail to compile. This was an intentional act to warn future dev about the change (imports and some other minor tweaks are necessary). 

## Linked Issues

- Closes #454


core/src/test_utils.rs
```diff
@@ -320,3 +320,116 @@ macro_rules! create_actors {
...
+            $evm_address,
+            $config.bridge_amount_sats,
+            $config.network,
+            $config.user_takes_after,
+        )
+        .0
+    }};
+}
+
+/// Gets the the deposit address for the user.
```
mmtftr, line 368: fix comment here

mmtftr (changes requested):
Looks like a simple changeset, let's make sure our documentation is good

PR #474, opened 2025-01-27
Implement key&script spend test on the same UTXO.

### LINKED ISSUE #427: Implement Test for Both Key and Script Spend Path with new MuSig2
### Issue Description

 We need to test the new MuSig2 with tweaking the Internal X-Only Public Key by a Script using the untweaked MuSig2 X-Only Public Key. We should be able spend using both ways.

### Tasks

Parts of the task:
- [ ] Implement a test that spends using both paths from a Taproot address.
- [ ] Document the test for future reference.

### Deadline of the Issue

-

mmtftr:
# Description

Verifies that a single P2TR with both key and script spend paths are actually able to be spent using the non-tweaked Musig2 aggregation **and** tweaked

Other changes:
- change default levelfilter to DEBUG (trace is way too aggressive in most cases)
- Use LevelFilter in internal API initialize_logger (did not change external-facing config signature)
- Added nice error message when a database connection can't be established in tests
- Refactored handle_taproot_witness_new to set_p2tr_script_spend_witness:
   - made it simpler, and removed incorrect Option. I believe the Option was used to make this function multi-purpose for both spend types, but this is a bad pattern as there is very little checking of the witness content in this case, and it's easy to misuse the function.
- Added set_p2tr_key_spend_witness, does a similar thing to above but for key spends and with type checks (by taking a valid taproot::Signature)

## Linked Issues

- Closes #427 

## Testing

This is a testing change with one new musig2 test.

### DIFF
--- a/core/src/test_utils.rs
+++ b/core/src/test_utils.rs
@@ -48,7 +48,7 @@ macro_rules! create_test_config_with_thread_name {
             + &suffix;
 
         // Use maximum log level for tests.
-        initialize_logger(5).unwrap();
+        initialize_logger(Some(::tracing::level_filters::LevelFilter::DEBUG)).unwrap();
 
         let mut config = BridgeConfig::default();
 
@@ -106,7 +106,12 @@ macro_rules! create_test_config_with_thread_name {
 macro_rules! initialize_database {
     ($config:expr) => {{
         let url = Database::get_postgresql_url(&$config);
-        let conn = sqlx::PgPool::connect(url.as_str()).await.unwrap();
+        let conn = sqlx::PgPool::connect(url.as_str())
+            .await
+            .expect(&format!(
+                "Failed to connect to database, please make sure a test Postgres DB is running at {}",
+                url
+            ));
 
         sqlx::query(&format!("DROP DATABASE IF EXISTS {}", &$config.db_name))
             .execute(&conn)
--- a/core/src/utils.rs
+++ b/core/src/utils.rs
@@ -2,10 +2,10 @@ use crate::builder::transaction::TxHandler;
 use crate::cli::Args;
 use crate::config::BridgeConfig;
 use crate::errors::BridgeError;
-use bitcoin;
-use bitcoin::sighash::SighashCache;
-use bitcoin::taproot::LeafVersion;
+use bitcoin::taproot::{self, LeafVersion};
 use bitcoin::XOnlyPublicKey;
+use bitcoin::{self, Witness};
+use tracing::Level;
 //use bitvm::chunker::assigner::BridgeAssigner;
 use std::borrow::BorrowMut;
 use std::collections::BTreeMap;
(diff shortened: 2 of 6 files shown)


core/src/utils.rs
```diff
@@ -115,32 +123,58 @@ pub fn usize_to_var_len_bytes(x: usize) -> Vec<u8> {
...
+/// # Arguments
+///
+/// - `tx`: The transaction to add the witness to.
+/// - `script_inputs`: The inputs to the tapscript
+/// - `txin_index`: The index of the transaction input to add the witness to.
+/// - `script_index`: The script index in the input UTXO's Taproot script tree. This is used to get the control block and script contents of the script being spent.
+pub fn set_p2tr_script_spend_witness<T: AsRef<[u8]>>(
     tx: &mut TxHandler,
-    witness_elements: &[T],
+    script_inputs: &[T],
```
ozankaymak, line 136: Rename to witness_elements maybe because Witness includes Script Inputs, Locking Script, and Control Block.

ozankaymak (changes requested):

core/src/test_utils.rs
```diff
@@ -48,7 +48,7 @@ macro_rules! create_test_config_with_thread_name {
             + &suffix;
 
         // Use maximum log level for tests.
-        initialize_logger(5).unwrap();
+        initialize_logger(Some(::tracing::level_filters::LevelFilter::DEBUG)).unwrap();
```
ceyhunsen, line 51: I guess you are right, this was a bit aggressive. But currently we have debug purposed log messages that uses `tracing::trace!()`. I am very skeptical about this change is worth it or not considering this.

core/src/utils.rs
mmtftr: But this input is only used for the script input. The locking script and control block are taken from the `taproot_spend_info`.

Thus, the original `witness_elements` naming is misleading, as it's not the entire witness contents.

core/src/test_utils.rs
mmtftr: For context, we only have 7 `trace` log statements in the codebase.

In addition, I don't think `trace` should ever be on by default globally. If a developer needs to debug these trace logs, they should enable them through `RUST_LOG`. Example:

```bash
# run header_chain_prover tests and enable tracing for that module only
RUST_LOG=debug,clementine_core::header_chain_prover=trace cargo test --package clementine-core --lib -- header_chain_prover::tests::new_with_proof_assumption --exact --show-output
```

In the worst case, if they want to debug everything with `trace` (low chance IMHO):
```bash
RUST_LOG=trace cargo test
```

ceyhunsen: Personally, I liked that every log is printed in case of an error and I can filter unnecessary logs after (that I am sure are not relevant). This behavior should be also useful for newcomers IMO (less setup to do).

Although my vote is still on trace level, I am fine with this change.

mmtftr: My beginner experience was like this:
1. I had a failing test but the logs were 100MB large and exceeded my terminal scrollback so I had no idea what was going wrong.
2. I pushed a branch and I couldn't even scroll to the error in Github Actions output because my browser couldn't load the entire log file.
3. I downloaded the logs archive, extracted and opened with vim to finally see the Error level log. 

I am in favor of this change unless there is a concrete argument against. I can't determine if this applies but this would be a valid argument:
> We have to see trace logs to debug in the majority of test failures

core/src/utils.rs
ozankaymak: Ah okay that makes more sense.

PR #485, opened 2025-01-28
Reorganize transaction builder

### LINKED ISSUE #482: Reorganize builder/transaction.rs into separate modules
### Proposal Description

The transaction builder module is too big and has too many different transactions.

I believe we have a couple of logical divisions: operator challenge sequence, operator sequential collateral sequence, operator reimburse, watchtower sequence. And we also have a couple of utilities mixed in that file, those should go out into another module.

The utilities should be moved into `util.rs`.

mmtftr:
# Description

Reorganizes code in builder/transaction into multiple submodules.

## Linked Issues

- Closes #482 

## Testing

No logic changes, just reorg


core/src/builder/address.rs
```diff
@@ -17,14 +17,20 @@ use bitcoin::{Amount, Network};
 use bitvm::signatures::winternitz;
 
 pub fn taproot_builder_with_scripts(scripts: &[ScriptBuf]) -> TaprootBuilder {
-    let n = scripts.len();
-    if n == 0 {
+    let num_scripts = scripts.len();
+
+    if num_scripts == 0 {
```
ceyhunsen, line 22: this function can be simplified with a mutable TaprootBuilder and 2 conditions that modifies it: num_scripts == 1 and num_scripts > 1. another improvement while we at it

core/src/builder/transaction/util.rs
ceyhunsen, line 1: id argue that these functions should stay in mod.rs

ceyhunsen (commented):
rest looks ok considering they are just simply moved to the new files

core/src/builder/transaction/util.rs
ozankaymak: Agreed

mmtftr: Done

core/src/builder/address.rs
mmtftr: Done, I wrote a test for it as well. Also, I found undocumented behavior in `rust-bitcoin`, using the same script multiple times will actually make it so that `script_map().len() != scripts.len()`

PR #486, opened 2025-01-28
Contextualize database and separate in to modules

### LINKED ISSUE #484: Split database/common into submodules
### Proposal Description

Database/common is currently very crowded and it needs to be contextualize.

ceyhunsen:
# Description

Separates database/common into different files based on the context.

This PR basically moves code to separate modules but it also removes `tracing::instrument`s (because they are useless) and adds some todos regarding to tests etc..

Adds execute_query_with_tx macro that will simplify query execution.

## Linked Issues

- Closes #484


mmtftr:
If you have the time, consider using this (I had written this earlier):

```rust
macro_rules! execute_with_tx {
    ($conn:expr, $tx:expr, $query:expr, $method:ident) => {
        match $tx {
            Some(tx) => $query.$method(&mut **tx).await,
            None => $query.$method(&$conn).await,
        }
    };
}


// later
execute_with_tx!(self.connection, tx, query, execute); 
```

ceyhunsen:
> If you have the time, consider using this (I had written this earlier):
> 
> ```rust
> macro_rules! execute_with_tx {
>     ($conn:expr, $tx:expr, $query:expr, $method:ident) => {
>         match $tx {
>             Some(tx) => $query.$method(&mut **tx).await,
>             None => $query.$method(&$conn).await,
>         }
>     };
> }
> 
> 
> // later
> execute_with_tx!(self.connection, tx, query, execute); 
> ```

Let me do this in a chain PR and not congest this PR, as there is already too many changes.

core/src/database/verifier.rs
```diff
@@ -0,0 +1,416 @@
...
+
+        execute_query_with_tx!(self.connection, tx, query, execute)?;
+
+        Ok(())
+    }
+
+    /// Verifier: Save the deposit info to use later
+    pub async fn save_deposit_info(
+        &self,
+        tx: Option<&mut sqlx::Transaction<'_, Postgres>>,
```
mmtftr, line 162: I believe we can also get this simple change in:

can we make a `pub type OptionalTransaction<'a> = Option<...>` then use it everywhere? (I think you'll need the lifetime parameter)

core/src/database/watchtower.rs
```diff
@@ -0,0 +1,273 @@
...
+use crate::errors::BridgeError;
+use crate::execute_query_with_tx;
+use bitcoin::{ScriptBuf, XOnlyPublicKey};
+use bitvm::signatures::winternitz;
+use bitvm::signatures::winternitz::PublicKey as WinternitzPublicKey;
+use sqlx::Postgres;
+
+impl Database {
+    /// Sets Winternitz public keys for an operator.
+    pub async fn save_watchtower_winternitz_public_keys(
```
mmtftr, line 15: Can we clarify the action of the function? "save" is ambiguously used for both insert and upsert (ie. insert if not exists else update). This one is an insert, so rename to `insert_watchtower_winternitz_public_keys`.

core/src/database/verifier.rs
```diff
@@ -0,0 +1,416 @@
...
+                    outpoint: utxo_db.0.outpoint_db.0,
+                    txout: utxo_db.0.txout_db.0.clone(),
+                })
+                .collect();
+            Ok(Some(utxos))
+        }
+    }
+
+    /// Verifier: Save the kickoff UTXOs for this deposit UTXO.
+    pub async fn save_kickoff_utxos(
```
mmtftr, line 78: Continuing from the other comment, this one is still an `insert`

core/src/database/watchtower.rs
```diff
@@ -0,0 +1,273 @@
...
+        let wpks: (Vec<u8>,) = execute_query_with_tx!(self.connection, tx, query, fetch_one)?;
+
+        let watchtower_winternitz_public_keys: Vec<winternitz::PublicKey> =
+            borsh::from_slice(&wpks.0).map_err(BridgeError::BorshError)?;
+
+        Ok(watchtower_winternitz_public_keys)
+    }
+
+    // TODO: Document
+    pub async fn save_watchtower_challenge_addresses(
```
mmtftr, line 58: continuing from before, this one is an `upsert`

mmtftr (changes requested):
Thanks a lot for the changes, I think we'll improve in terms of documentation with these small changes

ceyhunsen:
Let me make this changes in a chain PR again for an easier review experience. We should keep this PR clean from refactors/updates. I made a mistake merging the child PR #488.

After every chain PR is approved, we can merge them one by one. If we don't do this, some changes will be missed by the reviewers.

core/src/database/verifier.rs
ceyhunsen: Out of context for this PR. Conversation moved to #489

core/src/database/verifier.rs
ceyhunsen: Out of context for this PR. Conversation moved to #489

core/src/database/watchtower.rs
ceyhunsen: Out of context for this PR. Conversation moved to #489

ozankaymak (approved):

Issue #460, opened 2025-01-23
Pipeline for aggregator setup


ekrembal:
Instead of getting all params first and then passing them all, we should pass the params when we get it with something like queue. Just like we did in aggregator new_deposit.

Improves #455

doscortados:
Closed by https://github.com/chainwayxyz/clementine/pull/487

Issue #469, opened 2025-01-24
Clean-up RPC module from business logic and code repetition (for encoding/decoding)


ekrembal:
Right now, we need to do create_actors! then call aggregator.setup before doing any tests. This slows down our testing. Instead, every test should be able to spawn an entity to do it's job **without** creating a gRPC server.

PR #487, opened 2025-01-28
Aggregator setup pipelining

doscortados:
Closes https://github.com/chainwayxyz/clementine/issues/460


core/src/rpc/aggregator.rs
```diff
@@ -284,117 +286,122 @@ impl Aggregator {
     }
 }
 
+async fn collect<C, T, E, Fut, F>(clients: &[C], mut f: F) -> Result<Vec<T>, E>
```
mmtftr, line 289: Thanks a lot, these will reduce boilerplate by a lot.

I think the naming needs to change though, this seems like a utility function that calls all clients with a given func and returns the inner response. Something like: `map_clients_collect`?

Also maybe `#[inline]` and/or `#[track_caller]` are appropriate here.

core/src/rpc/aggregator.rs
```diff
@@ -284,117 +286,122 @@ impl Aggregator {
...
+{
+    try_join_all(
+        clients.iter().map(|client| {
+            f(client.clone()).map(|result| result.map(|response| response.into_inner()))
+        }),
+    )
+    .await
+}
+
+async fn pull<T>(mut stream: Streaming<T>) -> Result<Vec<T>, Status> {
```
mmtftr, line 303: I think we need docs on the behavior. This currently collects until end of stream unless an error happens, in which case we prematurely return the error. I think `collect` fits better here. This is also the behavior of `collect` when doing a `Iter<Item = Result<T, E>> -> Result<Vec<T>, E>`

Also can we move these two into another module?

mmtftr (changes requested):
Thanks a lot for the changes.

I believe the original issue also wants the `try_join_all`s to happen in background tasks so that they're happening in parallel (you can check aggregator for details)

We're also going to be working on [moving the business logic out of the gRPC handlers](https://github.com/chainwayxyz/clementine/issues/469), and moving parsing/validation logic out into trait impls so that they're more easily tested and the code has less clutter.

core/src/rpc/aggregator.rs
doscortados: Yes, this definitely needs refining, thank you for pointing it out. I didn't find an elegant way to generalize it over streaming (returning `Streaming<T>`) and regular/non-streaming (returning `Response<T>`) requests 🤷 .

core/src/rpc/aggregator.rs
doscortados: Thank you, this was just a first attempt to generalize stream collection, probably just `.collect()` on a stream would have been enough and this function is unnecessary.

doscortados:
Thank you for the comments @mmtftr , and sorry for my late reply!
This is rather initial version and significant re-work will follow in further commits.

At this point I see that `mpsc` channel isn't working out for multiplexing N-over-M streams, and I am trying out `mpmc` one to avoid pulling all the params in memory before streaming it to consumers.

core/src/rpc/aggregator.rs
```diff
@@ -393,116 +394,129 @@ impl ClementineAggregator for Aggregator {
...
-                Ok::<_, tonic::Status>(())
-            }
-        }))
-        .await?;
+        try_join_all([
+            operator_params_setup_task_handle,
+            watchtower_params_setup_task_handle,
+        ])
+        .await
+        .map_err(|e| BridgeError::Error(format!("aggregator setup failed: {e:?}")))?;
```
doscortados, line 519: Dear reviewer, 
Please suggest better error type here, this one seeks like a too general one.

ceyhunsen: Maybe `BridgeError::RPCStreamEndedUnexpectedly`? Or we can create a new one in `core/src/rpc/error.rs`.

ceyhunsen (changes requested):
Overall, it looks good. But apparently there is a deadlock or something like that: CI/CD tests are not finishing.

core/src/rpc/aggregator.rs
```diff
@@ -393,116 +394,129 @@ impl ClementineAggregator for Aggregator {
...
-
-                    for param in params {
-                        tx.send(param)
-                            .await
-                            .map_err(|_| output_stream_ended_prematurely())?;
-                    }
-
-                    future.await?; // TODO: This is dangerous: If channel size becomes not sufficient, this will block forever.
+        let operator_params_setup_task_handle = tokio::spawn(async move {
+            for response in operator_params_response_streams.iter_mut() {
```
mmtftr, line 449: Currently this loop:
1. For each operator
2. Stream all messages to operator (await)

This will block for each operator and won't pipeline operators in parallel.

can we collect all async blocks on the inside and join them on the outside of this loop?

core/src/rpc/aggregator.rs
```diff
@@ -393,116 +394,129 @@ impl ClementineAggregator for Aggregator {
...
 
-                    let future =
-                        client.set_watchtower(tokio_stream::wrappers::ReceiverStream::new(rx));
-                    for param in params {
-                        tx.send(param)
-                            .await
-                            .map_err(|_| output_stream_ended_prematurely())?;
+        let watchtower_params_setup_task_handle =
+            tokio::spawn(async move {
+                for response in watchtower_params_response_streams.iter_mut() {
```
mmtftr, line 494: same as above, let's collect the inner blocks and join on the outside

mmtftr (changes requested):

doscortados:
Thank you for your comments @mmtftr @ceyhunsen , I will address them once I get the tests pass.
Right now I figured out a "quick & dirty" way to avoid deadlock when broadcasting operator/watchtower params, but now when deadlock is no more, two tests still fail.

doscortados:
At this time the deadlock and the failing tests are fixed, I will make one more round of cleanups and mark it ready for the review.

core/src/rpc/aggregator.rs
```diff
@@ -386,106 +384,180 @@ impl ClementineAggregator for Aggregator {
...
-                            .map_err(|_| output_stream_ended_prematurely())?;
+                    let mut params = Vec::new();
+                    while let Some(param) = stream.message().await? {
+                        params.push(param);
                     }
-
-                    client
-                        .set_operator(tokio_stream::wrappers::ReceiverStream::new(rx))
-                        .await?;
+                    tx.send(StreamControl::Data(params)).map_err(|e| {
```
ceyhunsen, line 423: why not use Option instead?

core/src/rpc/aggregator.rs
```diff
@@ -386,106 +384,180 @@ impl ClementineAggregator for Aggregator {
...
+                                            "failed to propagate operator params: {e}"
+                                        ))
+                                    })?;
+                                }
+                                let _empty = stream.await?;
+                            }
+                            Ok(StreamControl::Done) => break Ok(()),
+                            Err(tokio::sync::broadcast::error::TryRecvError::Empty) => {
+                                // avoid race condition when producers did not yet produce anything
+                                tokio::time::sleep(std::time::Duration::from_millis(100)).await;
```
ceyhunsen, line 461: this must be avoided at all costs: not acceptable to sleep in async code. isnt this what async tries to solve?

ceyhunsen (changes requested):
It looks unnecessarily complex to me in general tbh

ekrembal (approved):

core/src/rpc/aggregator.rs
doscortados: Good point, no need to reinvent option!

core/src/rpc/aggregator.rs
doscortados: Why it is not acceptable to sleep in async code? This is "async" sleep, not blocking a thread or anything like that.

Also there is not much we can do about it - producers and consumers are spawned without any synchronization between them, so there might be a race condition when consumer is ready to consume but producer haven't yet produced anything.

doscortados:
> It looks unnecessarily complex to me in general tbh

I totally agree @ceyhunsen ! But it also is non-trivial multiplexing M-over-N streams in a concurrent fashion. Also it is necessary to re-stream from broadcast consumer to mpsc consumer due to limitation of the client API.

Compared with existing non-concurrent implementation the only overhead is actually only two extra broadcast channels (one for setting operator params, another one for watchtowers).

core/src/rpc/aggregator.rs
doscortados: Done https://github.com/chainwayxyz/clementine/pull/487/commits/1a86b4bfee4e0320791a853d7bfc5d3c9da3cf50

core/src/rpc/aggregator.rs
ceyhunsen: I don't have any solutions but would prefer to see an `.await` instead of sleeping while waiting other end of the stream. If you say there aren't any options for this, then it is what it is.

doscortados: `Condvar` might seem like a good fit here, then consumers can `.await` until the producers are actually producing something. I will try it out and post update here.

doscortados: When I think more about it, the possible race conditions in such dynamic setup (I mean multiplexing N-over-M streams) occur not only on startup, but also might occur "in flight" because of network lags - for example if due to network lags the producer slows down, faster consumers can consume all buffered messages and end up with the empty channel. In this case I don't think that much can be done about it - we literally need to **wait** for producers.

mmtftr: I don't get the necessity of this, this behavior with try_recv looks the same to me as just doing a recv. If producers did not produce anything, we'd just be awaiting on the recv

core/src/rpc/aggregator.rs
```diff
@@ -386,106 +384,174 @@ impl ClementineAggregator for Aggregator {
...
+        const CHANNEL_CAPACITY: usize = 1024 * 16;
+        let (operator_params_tx, _) = tokio::sync::broadcast::channel(CHANNEL_CAPACITY);
+
+        let tx = operator_params_tx.clone();
+        let clients = self.operator_clients.clone();
+        let get_operator_params_handle = tokio::spawn(async move {
+            tracing::info!(clients = clients.len(), "Collecting operator details...");
+            try_join_all(clients.iter().map(|client| {
+                let mut client = client.clone();
+                let tx = tx.clone();
```
mmtftr, line 406:
```suggestion
                let mut operator_client = client.clone();
```
for readability

core/src/rpc/aggregator.rs
```diff
@@ -386,106 +384,174 @@ impl ClementineAggregator for Aggregator {
...
+                    let mut params = Vec::new();
+                    while let Some(param) = stream.message().await? {
+                        params.push(param);
                     }
-
-                    client
-                        .set_operator(tokio_stream::wrappers::ReceiverStream::new(rx))
-                        .await?;
+                    tx.send(Some(params)).map_err(|e| {
+                        BridgeError::Error(format!("failed to forward operator params: {e}"))
```
mmtftr, line 418:
```suggestion
                        BridgeError::Error(format!("operator params stream closed unexpectedly: {e}"))
```

core/src/rpc/aggregator.rs
```diff
@@ -386,106 +384,174 @@ impl ClementineAggregator for Aggregator {
...
-            async move {
-                for params in operator_params {
-                    let (tx, rx) = tokio::sync::mpsc::channel(params.len());
+        // Propagate Operators configurations to all verifier clients
+        const CHANNEL_CAPACITY: usize = 1024 * 16;
+        let (operator_params_tx, _) = tokio::sync::broadcast::channel(CHANNEL_CAPACITY);
+
+        let tx = operator_params_tx.clone();
+        let clients = self.operator_clients.clone();
+        let get_operator_params_handle = tokio::spawn(async move {
```
mmtftr, line 402:
```suggestion
        let get_operator_params_chunked_handle = tokio::spawn(async move {
```

core/src/rpc/aggregator.rs
```diff
@@ -386,106 +384,174 @@ impl ClementineAggregator for Aggregator {
...
+            }))
+            .await?;
+            tx.send(None).map_err(|e| {
+                BridgeError::Error(format!("failed to close operator params stream: {e}"))
+            })?;
+            Ok::<_, Status>(())
+        });
+
+        let tx = operator_params_tx.clone();
+        let clients = self.verifier_clients.clone();
```
mmtftr, line 431:
```suggestion
        let verifiers = self.verifier_clients.clone();
```

core/src/rpc/aggregator.rs
```diff
@@ -386,106 +384,174 @@ impl ClementineAggregator for Aggregator {
...
+                            Ok(Some(params)) => {
+                                let (tx, rx) = channel(CHANNEL_CAPACITY);
+                                let stream = client.set_operator(ReceiverStream::new(rx));
+                                for param in params {
+                                    tx.send(param).await.map_err(|e| {
+                                        BridgeError::RPCStreamEndedUnexpectedly(format!(
+                                            "failed to propagate operator params: {e}"
+                                        ))
+                                    })?;
+                                }
```
mmtftr, line 449:
```suggestion
                                use futures::stream;
                                client.set_operator(stream::iter(params)).await?;
```
No need to create broadcast channels here, use Stream constructors. (also for the other calls where we're simply passing Vecs)

core/src/rpc/aggregator.rs
doscortados: ?

core/src/rpc/aggregator.rs
doscortados: Why chunked? It is a job that waits for all params.

core/src/rpc/aggregator.rs
mmtftr: I think we should rename `clients` to the actual entity types so that the code is easier to read. I don't know what `clients` is without finding its declaration (is it an operator? verifier? watchtower? etc.)

core/src/rpc/aggregator.rs
doscortados: Good one, thank you.

doscortados: (Just a follow up: `.await?` on `set_operator` seems to deadlock as well)

mmtftr: weird, what's the difference from what you're doing? you're still sending all of the params and waiting at the end.

core/src/rpc/aggregator.rs
doscortados: @mmtftr sorted out the issue, I only committed the code :)

core/src/rpc/aggregator.rs
doscortados: nevermind, the `stream::iter` works fine

core/src/rpc/aggregator.rs
doscortados: Nevermind, chunked it is.

ceyhunsen (approved):

core/src/rpc/aggregator.rs
```diff
@@ -390,106 +390,139 @@ impl ClementineAggregator for Aggregator {
...
+        let (watchtower_params_tx, watchtower_params_rx) =
+            tokio::sync::broadcast::channel(CHANNEL_CAPACITY);
 
-                let mut params_stream = client
-                    .get_params(Request::new(Empty {}))
-                    .await?
-                    .into_inner();
-                while let Some(response) = params_stream.message().await? {
-                    responses.push(response);
+        let tx = watchtower_params_tx.clone();
```
mmtftr, line 464: no need to clone and drop. just move the `watchtower_params_tx` inside

core/src/rpc/aggregator.rs
```diff
@@ -390,106 +390,139 @@ impl ClementineAggregator for Aggregator {
...
+        let verifiers = self.verifier_clients.clone();
+        let set_watchtower_params_handle = tokio::spawn(async move {
+            tracing::info!("Sending Winternitz public keys to verifiers...");
+            try_join_all(verifiers.iter().map(|verifier| {
+                let mut verifier = verifier.clone();
+                let watchtower_params_rx = &watchtower_params_rx;
+                async move {
+                    let mut watchtower_params_rx = watchtower_params_rx.resubscribe();
+                    loop {
+                        match watchtower_params_rx.recv().await {
```
mmtftr, line 497: can we extract this common logic into a function? Result<T, RecvError> -> Result<T, BridgeError>

mmtftr (commented):
Thanks a lot for the refactor, just some small comments

core/src/rpc/aggregator.rs
doscortados: done, `collect_and_call`

PR #488, opened 2025-01-29
Add execute_query_with_tx macro

ceyhunsen:
# Description

Adds a query executor with db tx support. See: https://github.com/chainwayxyz/clementine/pull/486#issuecomment-2620950369

### DIFF
--- a/core/src/database/mod.rs
+++ b/core/src/database/mod.rs
@@ -21,6 +21,24 @@ pub struct Database {
     connection: Pool<Postgres>,
 }
 
+/// Executes a query with a transaction if it is provided.
+///
+/// # Parameters
+///
+/// - `$conn`: Database connection.
+/// - `$tx`: Database transation, wrapped in `Option`.
+/// - `$query`: Query to execute.
+/// - `$method`: Method to execute on the query.
+#[macro_export]
+macro_rules! execute_query_with_tx {
+    ($conn:expr, $tx:expr, $query:expr, $method:ident) => {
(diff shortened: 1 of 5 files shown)


core/src/database/operator.rs
```diff
@@ -102,41 +91,33 @@ impl Database {
...
         );
 
-        let result: Result<Vec<(i32, String, String, TxidDB)>, sqlx::Error> = match tx {
-            Some(tx) => query.fetch_all(&mut **tx).await,
-            None => query.fetch_all(&self.connection).await,
-        };
+        let operators: Vec<(i32, String, String, TxidDB)> =
+            execute_query_with_tx!(self.connection, tx, query, fetch_all)?;
 
-        match result {
```
ceyhunsen, line 110: Only difference is that the result is handled with a ?

core/src/database/mod.rs
```diff
@@ -21,6 +21,24 @@ pub struct Database {
     connection: Pool<Postgres>,
 }
 
+/// Executes a query with a transaction if it is provided.
+///
+/// # Parameters
+///
+/// - `$conn`: Database connection.
+/// - `$tx`: Database transation, wrapped in `Option`.
```
mmtftr, line 29:
```suggestion
/// - `$tx`: Optional Database transaction
```

mmtftr (commented):
Looks good to me

mmtftr (approved):
Approved

PR #489, opened 2025-01-29
Update db function names, add tests and small improvements

ceyhunsen:
# Description

Adds small changes to the database module. Parent pr: #486 

- Added tests for db functions which lacks tests
- Updated namings of the functions
- Updated function comments

### DIFF
--- a/core/src/database/mod.rs
+++ b/core/src/database/mod.rs
@@ -21,6 +21,9 @@ pub struct Database {
     connection: Pool<Postgres>,
 }
 
+/// Optional database transaction.
+pub type DatabaseTransaction<'a, 'b> = Option<&'a mut sqlx::Transaction<'b, Postgres>>;
+
 /// Executes a query with a transaction if it is provided.
 ///
 /// # Parameters
--- a/core/src/database/mod.rs
+++ b/core/src/database/mod.rs
@@ -21,8 +21,8 @@ pub struct Database {
     connection: Pool<Postgres>,
 }
 
-/// Optional database transaction.
-pub type DatabaseTransaction<'a, 'b> = Option<&'a mut sqlx::Transaction<'b, Postgres>>;
+/// Database transaction for Postgres.
+pub type DatabaseTransaction<'a, 'b> = &'a mut sqlx::Transaction<'b, Postgres>;
 
 /// Executes a query with a transaction if it is provided.
 ///
--- a/core/src/database/verifier.rs
+++ b/core/src/database/verifier.rs
@@ -7,7 +7,7 @@ use super::{
         AddressDB, EVMAddressDB, MessageDB, MusigAggNonceDB, MusigPubNonceDB, OutPointDB,
         PublicKeyDB, TxOutDB, UtxoDB,
     },
-    Database,
+    Database, DatabaseTransaction,
 };
 use crate::{errors::BridgeError, execute_query_with_tx, EVMAddress, UTXO};
 use bitcoin::{
@@ -16,12 +16,12 @@ use bitcoin::{
     Address, OutPoint,
 };
 use secp256k1::musig::{MusigAggNonce, MusigPubNonce};
-use sqlx::{Postgres, QueryBuilder};
+use sqlx::QueryBuilder;
 
 impl Database {
     pub async fn save_verifier_public_keys(
         &self,
-        tx: Option<&mut sqlx::Transaction<'_, Postgres>>,
+        tx: DatabaseTransaction<'_, '_>,
         public_keys: &[PublicKey],
     ) -> Result<(), BridgeError> {
         let mut query = QueryBuilder::new("INSERT INTO verifier_public_keys (idx, public_key) ");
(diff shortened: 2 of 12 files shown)


ozankaymak (approved):

core/src/database/verifier.rs
```diff
@@ -16,19 +16,19 @@ use bitcoin::{
...
 };
 use secp256k1::musig::{MusigAggNonce, MusigPubNonce};
-use sqlx::{Postgres, QueryBuilder};
+use sqlx::QueryBuilder;
 
 impl Database {
-    pub async fn save_verifier_public_keys(
+    /// Sets the all verifiers' public keys. Given array **must** be in the same
+    /// order as the verifiers' indexes.
+    pub async fn set_verifiers_public_keys(
```
mmtftr, line 24: I still think these should be `insert` and `upsert`, inline with the SQL terminology, since these are a thin wrapper over SQL.

This function in particular is an `insert`.

core/src/database/mod.rs
```diff
@@ -21,6 +21,9 @@ pub struct Database {
     connection: Pool<Postgres>,
 }
 
+/// Optional database transaction.
+pub type DatabaseTransaction<'a, 'b> = Option<&'a mut sqlx::Transaction<'b, Postgres>>;
```
mmtftr, line 25: Can we name this OptionalTransaction? Otherwise, from the perspective of an outsider this would seem like a non-null transaction parameter that every db function requires (which it is not)

ozankaymak: Not to confuse with bitcoin::Transaction, maybe a different and more suitable name is better.

core/src/database/verifier.rs
ceyhunsen: Are you suggesting we should rename some "set"s to insert and upsert? Or just updating the docs? My approach was explaining the differences in the docs. If we change the prefix of the functions, it might be hard to find needed function because of an unknown prefix.

core/src/database/mod.rs
ceyhunsen: I didn't thought it will be that confusing for the developer, considering it's a one and only thing. Change wouldn't hurt also, though. I feel like naming is a bit off though but don't know why.

core/src/database/verifier.rs
mmtftr: I think we need to have consistency in naming and behavior regardless of docs. The docs are a great way to explain the internal behavior of a function, but the behavior of an insert on conflict is too important to be left in the docs (part of the core behavior IMO). The main benefit is when writing/reading code in the higher level actor classes, we'd like to have a good idea of the behavior of the db function without having to read the docs or the implementation

Specifically, if we're going to be using `set` for everything, then `set` should correspond to the same DB operation, either `insert` or `upsert`. Otherwise, it becomes confusing because `set` would be referring to a different operation in each case, and we'd have to check. (also, intuitively for me, `set` corresponds to the `update` SQL query, not `insert`. `set` sounds like you're modifying something that already exists, not adding something to the db) 

If we want to have different behavior on conflict, then we should have different names.

ceyhunsen: How about save_ for `insert`s and set_ for `upsert`s, then? I feel like using common words for these kind of functions are right thing to do. Because a caller doesn't care what is used behind the scenes as it only wants a specific operation is done. Heck, db module can even ditch SQL all together (don't know if there are alternatives though).

Using select_ instead of get_ sounds bad by the same approach, IMO.

I think that save_ can be a bit misleading too. It sounds like it can overwrite previous data. Open for suggestions.

mmtftr: I don't find the db interface that high-level at the moment, it's not agnostic of the underlying storage layer. 

- It takes a db transaction so the caller needs to understand db behavior.
- The caller also needs to understand the unique fields and the stored fields.

It's currently like an ORM function with parsing and decoding, so you still need to know what's going on underneath.

Also I can't discern the difference between `save` and `set` without reading the docs. I still think `insert` and `upsert` are more standard terminology. Even if we ditch SQL, the terminology will be correct

core/src/database/mod.rs
ceyhunsen: I think the solution here is https://github.com/chainwayxyz/clementine/pull/489/commits/862633219d91fe0460c87f75a8ad52346685fd99. With this, caller can be sure that db tx is not required and can just skip using a `None`. If a caller needs this, then they can check what is the type definition and act accordingly.

In the future, if we want to force a db function to be included in a db tx, we won't needto create a new type and not just wrap around an `Option`.

This also won't lengthen the parameter line in db functions; Close numbers of characters.

mmtftr: Ok, sounds good to me

mmtftr (approved):

core/src/database/verifier.rs
ceyhunsen: This conversation doesn't sound resolved but the sake of this PR getting merged, I moved the discussion to #491. We can open another PR after.

PR #490, opened 2025-01-29
Convert NofN key spends to checksig script spends

### LINKED ISSUE #461: Convert N-of-N key-path spends to script path spend.
Let's not try to do tweaks now. Adding the N-of-N to script spends helps us a lot, since Signer's don't have to apply any tweaks which has some computational costs. see: https://github.com/bitcoin-core/secp256k1/blob/f79f46c70386c693ff4e7aef0b9e7923ba284e56/src/modules/musig/keyagg_impl.h#L247

mmtftr:
# Description

This PR moves the keyspend to the script path to reduce work done by signers/aggregators with tweaking.

## Linked Issues

- Closes #461 

## Testing

None, we should add spending tests.

### DIFF
--- a/core/src/builder/address.rs
+++ b/core/src/builder/address.rs
@@ -147,11 +147,11 @@ pub fn generate_deposit_address(
 /// - [`Address`]: Checksig taproot Bitcoin address
 /// - [`TaprootSpendInfo`]: Checksig address's taproot spending information
 pub fn create_checksig_address(
-    nofn_xonly_pk: XOnlyPublicKey,
+    xonly_pk: XOnlyPublicKey,
     network: bitcoin::Network,
 ) -> (Address, TaprootSpendInfo) {
-    let nofn_script = builder::script::generate_checksig_script(nofn_xonly_pk);
-    create_taproot_address(&[nofn_script], None, network)
+    let script = builder::script::generate_checksig_script(xonly_pk);
+    create_taproot_address(&[script], None, network)
 }
 
 pub fn derive_challenge_address_from_xonlypk_and_wpk(
--- a/core/src/builder/address.rs
+++ b/core/src/builder/address.rs
@@ -130,7 +130,7 @@ pub fn generate_deposit_address(
     let recovery_extracted_xonly_pk =
         XOnlyPublicKey::from_slice(&recovery_script_pubkey.as_bytes()[2..34]).unwrap();
 
-    let script_timelock = builder::script::generate_relative_timelock_script(
+    let script_timelock = builder::script::generate_checksig_relative_timelock_script(
         recovery_extracted_xonly_pk,
         user_takes_after,
     );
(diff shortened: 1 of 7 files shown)


core/src/builder/address.rs
```diff
@@ -130,28 +130,28 @@ pub fn generate_deposit_address(
...
 /// # Returns
 ///
 /// See [`create_taproot_address`].
 ///
-/// - [`Address`]: MuSig2 taproot Bitcoin address
-/// - [`TaprootSpendInfo`]: MuSig2 address's taproot spending information
-pub fn create_musig2_address(
+/// - [`Address`]: Checksig taproot Bitcoin address
+/// - [`TaprootSpendInfo`]: Checksig address's taproot spending information
+pub fn create_checksig_address(
```
ozankaymak, line 149: nofn_checksig seems better

core/src/builder/script.rs
```diff
@@ -116,9 +132,20 @@ pub fn actor_with_preimage_script(
 }
 
 /// Generates a signature verification script.
-pub fn checksig_script(actor_taproot_xonly_pk: XOnlyPublicKey) -> ScriptBuf {
+///
+/// This script is a simple P2PKH script that pays to the signature generated from the
```
ozankaymak, line 136: Is this P2PKH?

core/src/builder/transaction/operator_assert.rs
```diff
@@ -144,9 +145,11 @@ pub fn create_assert_end_txhandler(
     });
 
     let disprove_taproot_spend_info = TaprootBuilder::new()
-        .add_hidden_node(0, TapNodeHash::from_slice(root_hash).unwrap())
+        .add_hidden_node(1, TapNodeHash::from_slice(root_hash).unwrap())
```
ozankaymak, line 148: Thanks for the changes, though I think we shouldn't create a P2TR address outside the given functionalities like this. We should change this in the future.

ozankaymak (changes requested):

core/src/builder/address.rs
mmtftr: this script is not limited to nofn, it could be any PK. Plus, there is no way of knowing that the pk is an nofn pk anyway, this function returns a generic p2pk script.

I'm strongly biased towards naming it by what it does, not how it's used downstream

mmtftr: I should also change the parameter name and variable namings

core/src/builder/script.rs
mmtftr: ~H~

core/src/builder/transaction/operator_assert.rs
mmtftr: reverted and added todo

core/src/builder/address.rs
ozankaymak: Then it should be fine

ozankaymak (approved):

core/src/builder/address.rs
mmtftr: Yeah, changed the naming so resolved.

ceyhunsen (approved):

PR #495, opened 2025-02-03
Add macros/parsers to RPC module

### LINKED ISSUE #441: Standardize gRPC parsing within business logic code
### Proposal Description

# Description
Currently, we have a lot of parsing and error handling code within our RPC handlers. An example is given below of one. Instead of including this parsing code within the business logic (i.e. our rpc handlers in this case), we should be writing wrappers that parse and convert while returning domain errors (BridgeError etc.).

We actually have a wrapper module that handles conversion with errors, but it's very limited at the moment. Below is an example of the conversion from our RPC OutPoint to bitcoin::OutPoint.

https://github.com/chainwayxyz/clementine/blob/1bfab0915c8b1b9a9cab40138e4422274346e56b/core/src/rpc/wrapper.rs#L8-L22

We should expand the functionality of this wrapper module to avoid writing parsing code inside the rpc handlers. Ideally, the below function should be hidden behind a `.try_into()?` (also a side note: u32s would be better off being a new type)

```rs

fn get_deposit_params(
    deposit_sign_session: clementine::DepositSignSession,
    verifier_idx: usize,
) -> Result<
    (
        bitcoin::OutPoint,
        EVMAddress,
        bitcoin::Address<NetworkUnchecked>,
        u32,
        u32,
    ),
    Status,
> {
    let deposit_params = deposit_sign_session
        .deposit_params
        .ok_or(Status::internal("No deposit outpoint received"))?;
    let deposit_outpoint: bitcoin::OutPoint = deposit_params
        .deposit_outpoint
        .ok_or(Status::internal("No deposit outpoint received"))?
        .try_into()?;
    let evm_address: EVMAddress = deposit_params.evm_address.try_into().unwrap();
    let recovery_taproot_address = deposit_params
        .recovery_taproot_address
        .parse::<bitcoin::Address<_>>()
        .map_err(|e| Status::internal(e.to_string()))?;
    let user_takes_after = deposit_params.user_takes_after;

    let session_id = deposit_sign_session.nonce_gen_first_responses[verifier_idx].id;
    Ok((
        deposit_outpoint,
        evm_address,
        recovery_taproot_address,
        user_takes_after,
        session_id,
    ))
}
```

ceyhunsen:
# Description

Adds new macros/parsers that simplifies RPC module. This PR will move parsing code that are mostly for streams, in to a separate module called `parser`.

## Linked Issues

- Closes #441


core/src/rpc/parser/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+use std::num::TryFromIntError;
+use tonic::Status;
+
+pub mod operator;
+pub mod verifier;
+pub mod watchtower;
+
+/// Converts an integer type in to another integer type. This is needed because
+/// tonic defaults to wrong integer types for some parameters.
+pub fn convert_int_to_another<SOURCE, TARGET>(
```
ekrembal, line 15: Do you think if it helps making this a macro?

ekrembal (approved):

core/src/rpc/parser/mod.rs
ceyhunsen: I like the functions better than macros, just because ease of use. In this case, it is possible to make this a function. So, I made it a function.

I think specifying trait bounds is another benefit. But I couldn't find a specific trait bounds for integer types. Don't think a macro will be more limiting to the inputs.

ceyhunsen:
Figured I could add other improvements. Switching to draft...

core/src/rpc/parser/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+/// // A response enum, that includes possible stream elements. This will panic
+/// // if the stream has ended.
+/// let operator_param: operator_params::Response = fetch_next_from_stream!(stream, response, "response").unwrap();
+///
+/// // A response enum wrapped around by an [`Option`], that includes possible
+/// // stream elements. This won't panic if stream has ended.
+/// let operator_param: Option<operator_params::Response> = fetch_next_from_stream!(stream, response);
+/// ```
+#[macro_export]
+macro_rules! fetch_next_message_from_stream {
```
mmtftr, line 58: I think the return value should be the same for a function/macro regardless of arguments. Can we please make a secondary macro for fetch_next_from_stream_opt?

Also, can we please put a default string for the expected_msg_got_none field argument that is simply the field name stringified? (You can use [`stringified!`](https://doc.rust-lang.org/std/macro.stringify.html), if possible it'd be nice to have `stringify!($stream.$field)`

ceyhunsen: I looked up something like stringify but couldn't find, thanks!

core/src/rpc/parser/operator.rs
```diff
@@ -0,0 +1,100 @@
...
+    stream: &mut tonic::Streaming<OperatorParams>,
+) -> Result<(u32, Txid, XOnlyPublicKey, Address), Status> {
+    let operator_param = fetch_next_message_from_stream!(stream, response, "response")?;
+
+    let operator_config =
+        if let operator_params::Response::OperatorDetails(operator_config) = operator_param {
+            operator_config
+        } else {
+            return Err(expected_msg_got_none("OperatorDetails")());
+        };
```
mmtftr, line 33:
```suggestion
      let operator_params::Response::OperatorDetails(operator_config) = operator_param else {
            return Err(expected_msg_got_none("OperatorDetails")());
       };
```

mmtftr, line 33:
```suggestion
      let operator_params::Response::OperatorDetails(operator_config) = operator_param else {
            return Err(expected_msg_got_none("OperatorDetails")());
       };
```

core/src/rpc/parser/verifier.rs
```diff
@@ -0,0 +1,120 @@
...
+                error::invalid_argument(
+                    "partial_sig",
+                    format!("Verifier {idx} returned an invalid partial signature").as_str(),
+                )(e)
+            })
+        })
+        .collect::<Result<Vec<_>, _>>()
+}
+
+pub async fn parse_next_deposit_finalize_param_schnorr_sig(
```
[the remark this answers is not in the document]
ceyhunsen: I guess so. But doesn't it look a bit misleading? I mean it is too much operations for a `try_into`. It just isn't converting some type to another but reading a stream and then converting them.

Currently, I am adding more `TryFrom` trait impls. I can also write trait impls which are for the inside of `parse_` functions. Let me do that and we can evaluate this again.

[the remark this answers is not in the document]
mmtftr: nvm, tryfrom can't be async anyway and you have a point. also, if the function is doing too many things, it should not do all of the things together, we should split them into separate functions. can we also shorten the naming?

core/src/rpc/aggregator.rs
```diff
@@ -235,30 +235,10 @@ async fn create_nonce_streams(
     }))
     .await?;
 
-    // Get the first responses.
+    // Get the first responses from verifiers.
     let first_responses: Vec<clementine::NonceGenFirstResponse> =
         try_join_all(nonce_streams.iter_mut().map(|stream| async {
```
mmtftr, line 240:
```suggestion
        try_join_all(nonce_streams.iter_mut().map(parser::verifier::parse_nonce_gen_first_response)
```

not sure if it compiles but

core/src/rpc/parser/mod.rs
ceyhunsen: https://github.com/chainwayxyz/clementine/pull/495/commits/1fe42a08157c9aab7252f362293c2b47e6b68c74

core/src/rpc/aggregator.rs
ceyhunsen: lets do this in another pr: out of context and this pr needs to be merged quickly

core/src/rpc/parser/verifier.rs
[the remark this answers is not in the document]
ceyhunsen: I think I can only remove `parse_` prefix from names. Though they can stay. Open for better naming suggestions.

mmtftr (approved):
Looks good to me, thanks a lot for this refactor

PR #498, opened 2025-02-04
Calculate BitVM with dummy winternitz at init

ekrembal:


core/src/utils.rs
```diff
@@ -41,24 +55,220 @@ lazy_static::lazy_static! {
...
+#[derive(BorshSerialize, BorshDeserialize, Debug)]
+pub struct BitvmCache {
+    pub intermediate_variables: BTreeMap<String, usize>,
+    pub disprove_scripts: Vec<Vec<u8>>,
+    pub replacement_places: HashMap<(usize, usize), Vec<(usize, usize)>>,
+}
+
+#[cfg(not(debug_assertions))]
+impl BitvmCache {
+    fn save_to_file(&self, path: &str) -> bool {
```
ceyhunsen, line 125: why not return result instead?

core/src/utils.rs
```diff
@@ -41,24 +55,220 @@ lazy_static::lazy_static! {
...
+            }
+
+            let winternitz_pk = WinternitzPublicKey {
+                public_key: winternitz_pk,
+                parameters,
+            };
+            Ok((intermediate_step.clone(), winternitz_pk))
+        })
+        .collect::<Result<BTreeMap<_, _>, BridgeError>>()
+        .unwrap();
```
ceyhunsen, line 192: this looks like infallible. why return result?

ceyhunsen:
Does release build test has a deadlock or new BitVM code is this slow when enabled? At the time of writing, debug tests ended in 4m while release tests are still running after 15m.

core/src/utils.rs
ekrembal: Fixed in 472d108

core/src/utils.rs
```diff
@@ -41,24 +55,220 @@ lazy_static::lazy_static! {
...
+                }
+            },
+            Err(e) => {
+                tracing::error!("Failed to serialize BitVM cache: {}", e);
+                false
+            }
+        }
+    }
+
+    fn load_from_file(path: &str) -> Option<Self> {
```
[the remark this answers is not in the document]
ekrembal: Fixed in 472d108

core/src/utils.rs
ekrembal: Fixed in 472d108

ceyhunsen (approved):
LGTM. Still needs tests though.

ekrembal:
We can add tests later, this is not the final way we will use BitVM anyways

Issue #175, opened 2024-05-28
Configuration File Documentation And Structure Updates


ceyhunsen:
### Proposal Description

Currently, there are only test configuration files and internal struct documentation provided. For the time being, this is OK. But in the feature, this will cause obvious problems. There 2 things to be done to solve this:

## Documentation

Clementine should have a user friendly documentation for configuration files, like markdown format. This can sit in `docs` directory.

## Sections

There is also one small obstacle against user friendliness: No sections. Currently configuration file looks very complicated because of this. Take this for example:

```toml
user_takes_after = 200
port = 3000
db_password = ""
secret_key = "5555555555555555555555555555555555555555555555555555555555555555"
confirmation_treshold = 1
db_port = 5432
verifiers_public_keys = [
    "4f355bdcb7cc0af728ef3cceb9615d90684bb5b2ca5f859ab0f0b704075871aa",
    "466d7fcae563e5cb09a0d1870bb580344804617879a14949cf22285f1bae3f27",
    "3c72addb4fdf09af94f0c94d7fe92a386a7e70cf8a1d85916386bb2535c7b1b1",
    "2c0b7cf95324a07d05398b240174dc0c2be444d96b159aa6c7f7b1e668680991",
    "9ac20335eb38768d2052be1dbbc3c8f6178407458e51e6b4ad22f1d91758895b",
]
db_user = "clementine"
num_verifiers = 4
tracing_debug = "debug,bitcoincore_rpc=info,hyper=error"
min_relay_fee = 305
host = "127.0.0.1"
network = "regtest"
bitcoin_rpc_url = "http://127.0.0.1:18443"
db_name = "clementine"
bitcoin_rpc_password = "admin"
all_secret_keys = [
    "1111111111111111111111111111111111111111111111111111111111111111",
    "2222222222222222222222222222222222222222222222222222222222222222",
    "3333333333333333333333333333333333333333333333333333333333333333",
    "4444444444444444444444444444444444444444444444444444444444444444",
    "5555555555555555555555555555555555555555555555555555555555555555",
]
db_host = "127.0.0.1"
bitcoin_rpc_user = "admin"
```

This example is intentionally mixed to show it can get very complicated. Even not intentionally modified, it can get still get this confusing by any cause.

To solve this issue, sections can be used. Same example with sections:

```toml
tracing_debug = "debug,bitcoincore_rpc=info,hyper=error"

num_verifiers = 4
min_relay_fee = 305
user_takes_after = 200
[...]

PR #500, opened 2025-02-05
RPC module de business logicification

### LINKED ISSUE #515: Move business logic code outside of rpc module
### Proposal Description

- [ ] Use regular `entity::do_a_thing()` calls to generate return data of a function
- [ ] Take advantage of `tokio` channels for the stream typed inputs/outputs

ceyhunsen:
# Description

This PR moves out business logic code from RPC module.

- Adds another transportation layer (tokio mpsc channels) between actor modules and rpc module (only for streams) in order to achieve this
- Maintains async nature of the streams
- Only required changes to the actual code apart from moving the code
- Tests are untouched to ensure previous working state is preserved

## Linked Issues

- Related #469  
- Closes #515

### DIFF
--- a/core/src/operator.rs
+++ b/core/src/operator.rs
@@ -118,6 +118,45 @@ impl Operator {
         })
     }
 
+    /// Returns an operator's winternitz public keys and challenge ackpreimages
+    /// & hashes.
+    ///
+    /// # Returns
+    ///
+    /// - [`mpsc::Receiver`]: A [`tokio`] data channel with a type of
+    ///   [`winternitz::PublicKey`] and size of operator's winternitz public
+    ///   keys count
+    /// - [`mpsc::Receiver`]: A [`tokio`] data channel with a type of
+    ///   [`PublicHash`] and size of operator's challenge ack preimages & hashes
+    ///   count
+    pub async fn get_params(
+        &self,
+    ) -> Result<
+        (
+            mpsc::Receiver<winternitz::PublicKey>,
+            mpsc::Receiver<PublicHash>,
+        ),
+        BridgeError,
+    > {
+        let wpks = self.get_winternitz_public_keys()?;
+        let wpk_channel = mpsc::channel(wpks.len());
+
+        let hashes = self.generate_challenge_ack_preimages_and_hashes()?;
+        let hashes_channel = mpsc::channel(hashes.len());
+
+        tokio::spawn(async move {
+            for wpk in wpks {
+                wpk_channel.0.send(wpk).await.expect("Can't send wpk"); // TODO: Will be handled
+            }
+
+            for hash in hashes {
+                hashes_channel.0.send(hash).await.expect("Can't send hash"); // TODO: Will be handled
+            }
+        });
+
+        Ok((wpk_channel.1, hashes_channel.1))
+    }
+
     pub async fn deposit_sign(
         &self,
         deposit_outpoint: OutPoint,
--- a/core/src/operator.rs
+++ b/core/src/operator.rs
@@ -146,12 +146,20 @@ impl Operator {
 
         tokio::spawn(async move {
             for wpk in wpks {
-                wpk_channel.0.send(wpk).await.expect("Can't send wpk"); // TODO: Will be handled
+                wpk_channel
+                    .0
+                    .send(wpk)
+                    .await
+                    .map_err(|e| BridgeError::SendError("winternitz public key", e.to_string()))?;
             }
 
             for hash in hashes {
-                hashes_channel.0.send(hash).await.expect("Can't send hash"); // TODO: Will be handled
+                hashes_channel.0.send(hash).await.map_err(|e| {
+                    BridgeError::SendError("challenge_ack_preimages_and_hashes", e.to_string())
+                })?;
             }
+
+            Ok::<(), BridgeError>(())
         });
 
         Ok((wpk_channel.1, hashes_channel.1))
@@ -188,15 +196,14 @@ impl Operator {
         tokio::spawn(async move {
             while let Some(sighash) = sighash_stream.next().await {
                 // None because utxos that operators need to sign do not have scripts
-                let sig = operator
-                    .signer
-                    .sign_with_tweak(sighash.expect("Expected sig").0, None)
-                    .expect("Error while sigbing with tweak");
+                let sig = operator.signer.sign_with_tweak(sighash?.0, None)?;
 
                 if sig_tx.send(sig).await.is_err() {
                     break;
                 }
             }
+
+            Ok::<(), BridgeError>(())
         });
 
         Ok(sig_rx)
--- a/core/src/operator.rs
+++ b/core/src/operator.rs
@@ -986,4 +986,32 @@ mod tests {
                 * config.num_watchtowers
         );
     }
+
+    #[tokio::test]
+    async fn operator_get_params() {
+        let config = create_test_config_with_thread_name!(None);
+        let rpc = ExtendedRpc::connect(
+            config.bitcoin_rpc_url.clone(),
+            config.bitcoin_rpc_user.clone(),
+            config.bitcoin_rpc_password.clone(),
+        )
+        .await
+        .unwrap();
+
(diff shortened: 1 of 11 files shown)


core/src/verifier.rs
```diff
@@ -114,6 +136,542 @@ impl Verifier {
...
+        Ok(())
+    }
+
+    pub async fn set_operator(
+        &self,
+        operator_idx: u32,
+        operator_xonly_pk: XOnlyPublicKey,
+        wallet_reimburse_address: Address,
+        collateral_funding_txid: Txid,
+        operator_winternitz_public_keys: Vec<winternitz::PublicKey>,
```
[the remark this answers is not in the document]
ceyhunsen: yes, you are right. i was trying to do the same but encountered deadlocks. so, i did commit temporary fixes. it looks like this pr will improve code quality but reduce performance because it will introduce another transportation layer. i will open up a discussion about this.

core/src/operator.rs
```diff
@@ -114,6 +118,90 @@ impl Operator {
...
+        ));
+
+        let operator = self.clone();
+        tokio::spawn(async move {
+            while let Some(sighash) = sighash_stream.next().await {
+                // None because utxos that operators need to sign do not have scripts
+                let sig = operator
+                    .signer
+                    .sign_with_tweak(sighash.expect("Expected sig").0, None)
+                    .expect("Error while sigbing with tweak");
```
mmtftr, line 194: why are these panics?

these look very much like recoverable errors. Please return recoverable errors here.

core/src/operator.rs
```diff
@@ -423,7 +520,19 @@ impl Operator {
             }
         }
 
-        if !self.is_profitable(input_utxo.txout.value, output_txout.value)? {
+        let operator_withdrawal_fee_sats =
+            self.config
+                .operator_withdrawal_fee_sats
+                .ok_or(BridgeError::ConfigError(
```
mmtftr, line 526: config errors should be handled on start. We should have a valid config while running.

If possible, the `self.config` should be statically known to be valid (ie. not an option or a result)

mmtftr: this might be out of scope for this PR

core/src/rpc/operator.rs
```diff
@@ -61,48 +59,30 @@ impl ClementineOperator for Operator {
...
-        });
+        while let Some(sig) = deposit_signatures_rx.recv().await {
+            let operator_burn_sig = OperatorBurnSig {
+                schnorr_sig: sig.serialize().to_vec(),
+            };
+
+            if tx.send(Ok(operator_burn_sig)).await.is_err() {
+                break;
+            }
+        }
```
mmtftr, line 85: I believe it'd be better to do
```rust
Ok(Response::new(ReceiverStream::new(deposit_signatures_rx).map(|sig| OperatorBurnSig {
  schnorr_sig: sig.serialize().to_vec()
})))
```

core/src/rpc/watchtower.rs
```diff
@@ -15,33 +15,29 @@ impl ClementineWatchtower for Watchtower {
...
         _request: Request<Empty>,
     ) -> Result<Response<Self::GetParamsStream>, Status> {
-        let watchtower = self.clone();
-        let watchtower_winternitz_public_keys =
-            watchtower.get_watchtower_winternitz_public_keys().await?;
+        let (watchtower_id, mut winternitz_public_keys, xonly_pk) = self.get_params().await?;
 
-        let (tx, rx) = mpsc::channel(watchtower_winternitz_public_keys.len() + 2);
+        let (tx, rx) = mpsc::channel(winternitz_public_keys.max_capacity() + 2);
         let out_stream: Self::GetParamsStream = ReceiverStream::new(rx);
```
mmtftr, line 21: this is more readable (though keep in mind, it has different semantics, it won't fill up the channel beforehand like the `tokio::spawn` will and won't be executed in the background, [ie. it acts like any old Future]).
```rust
let stream = futures::stream::iter(&[WatchTowerId]) // the long conversion in L30
  .chain(ReceiverStream::from(winternitz_public_keys).map(Into::into))
  .chain(futures::stream::iter(&[xonly_pk.into()]));
Ok(Response::new(stream))
```
you may need to add more type hints, to coerce it to be a stream of WatchtowerParams

core/src/watchtower.rs
```diff
@@ -99,6 +100,34 @@ impl Watchtower {
...
+        &self,
+    ) -> Result<(u32, mpsc::Receiver<winternitz::PublicKey>, XOnlyPublicKey), BridgeError> {
+        let watchtower_id = self.config.index;
+        let winternitz_public_keys = self.get_watchtower_winternitz_public_keys().await?;
+        let xonly_pk = self.actor.xonly_public_key;
+
+        let (wpk_channel_tx, wpk_channel_rx) = mpsc::channel(winternitz_public_keys.len());
+
+        tokio::spawn(async move {
+            for wpk in winternitz_public_keys {
```
mmtftr, line 122: I believe the following happens here:

This background task will complete immediately, since the buffer length == total items sent.

If that's the case, this doesn't need to be a background task, simply send the entire vec into the channel within the functino. You know for a fact that it won't block. Since the tx is dropped, the rx will drop values as they're consumed.

In fact, the semantics would be even better if you returned:

```rust
ReceiverStream::from(stream::iter(winternitz_public_keys))
```
which will likely have the same effect but gets rid of the extra background task and all the confusing error handling (which is nothing but handling the closure and early dropping of each stream)

mmtftr (commented):
I believe we can write the stream operations more clearly. I've added some improvements. Some may be out of scope but I believe some others are much needed simplification of stream handling in our codebase.

core/src/operator.rs
ceyhunsen: All of the new expects should be inside `tokio::spawn`s. Because we didn't decided how we will handle this errors yet, I just left them as is. I think it is better to convert them while working on #502 because `Result`s will be discarded anyways rn. For today, we can identify which line returned an error while testing.

core/src/operator.rs
ceyhunsen: We use a single config struct for every entity and some entities have config options which others don't. And we don't have any way to tell which entity is running. That's why some of the functions have to do these checks.

If we decided to implement #175, we can have different entity configurations without duplicating codebase (new issue required or existing one can be updated).

Also this code is mostly untouched. I just changed `is_profitable`'s return type into `bool` from `Result<bool, ...>`. So this is out of scope.

core/src/rpc/watchtower.rs
ceyhunsen: Good catch. This approach is only useful if calculations can be made async but in this case it is serial. Fixed in https://github.com/chainwayxyz/clementine/pull/500/commits/22a9b2c461e024799068a62712f9b8864eca6bb3

core/src/watchtower.rs
ceyhunsen: Good catch. This approach is only useful if calculations can be made async but in this case it is serial. Fixed in https://github.com/chainwayxyz/clementine/pull/500/commits/22a9b2c461e024799068a62712f9b8864eca6bb3

core/src/operator.rs
mmtftr: OK that it's out of scope. We really should have enums for the config in this case, or parse on start for the entity and store the validated information inside the entity struct

core/src/operator.rs
ceyhunsen: Yeah, you are right. Fixed in https://github.com/chainwayxyz/clementine/pull/500/commits/0b5f52b4a52721f520b24f2de3b58acc4a06880c.

core/src/rpc/operator.rs
ceyhunsen: It is discussed to push this change to a future PR, in a meeting.

core/src/rpc/watchtower.rs
ceyhunsen: It is discussed to push this change to a future PR, in a meeting.

ekrembal (approved):

mmtftr (approved):

PR #508, opened 2025-02-06
Save signatures with more information to db

### LINKED ISSUE #503: Improve signature saving to database
### Proposal Description

Instead of putting signatures in a flat array, we can do it in smarter way.

New tables:

- deposit_outpoints: only for indexing deposits. Columns: deposit_id, deposit_outpoint
- deposit_signatures_per_kickoff: columns: deposit_id, operator_id, sequential_collateral_idx, kickoff_connector_idx, signatures

We should store signatures in a good way, we can use protobuf with something like:

```protobuf
enum SignatureType {
  OperatorTimeoutTx = 0;
  OperatorTakeTx = 1;
  OperatorReimburseTx = 2;
  OperatorChallengeNackFirst = 3;
  OperatorChallengeNackSecond = 4;
}

message Signature {
  SignatureType signature_type = 1;
  int32 watctower_idx = 2;
  bytes signature = 3;
}

message DepositSignatures { repeated Signature signatures = 1; }
```

Also, since we store signatures in a structured way, we should be able to fill signatures of a TxHandler, with calling TxHandler.FillSignatures(&SignaturesObject)

atacann:
# Description
-While creating TxHandler add information about spend path of input (keypath or script path idx), tag every input with an enum so that we can easily understand which saved nofn signature corresponds to which utxo. 
-Save signatures with their tag to db, each row in signatures db is denoted by deposit_outpoint (converted to a unique int), operator_idx, sequential_collateral_idx and kickoff_idx.

## Linked Issues

- Closes #503

### DIFF
--- a/core/src/actor.rs
+++ b/core/src/actor.rs
@@ -299,6 +299,7 @@ mod tests {
     use crate::builder::transaction::output::UnspentTxOut;
     use crate::builder::transaction::TxHandlerBuilder;
     use crate::config::BridgeConfig;
+    use crate::rpc::clementine::NormalSignatureKind;
     use crate::utils::{initialize_logger, SECP};
     use crate::{
         actor::WinternitzDerivationPath, builder::transaction::TxHandler,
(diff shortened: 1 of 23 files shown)


atacann:
> LGTM. But can't we rename NormalSignature to OperatorSignature or NofNSignature, if I am not missing something? NormalSignature doesn't ring a bell for me TBH.

Yes NormalSignature is not descriptive enough. In this case NormalSignature can include both operator and NofN sigs. It means that there is only one signature per one kickoff, but the other type (WatchtowerSignature) has sigs for each watchtower so they need an additional watchtower_idx parameter.

mmtftr:
Renaming to `SingularSignature`

PR #509, opened 2025-02-06
Tx Sender

### LINKED ISSUE #465: Simple TxSender
Simple TxSender for regtest, it should accept Raw transactions and send it to Bitcon RPC. If V3 transactions are not accepted, add a child transaction.

### LINKED ISSUE #520: Improve Tx Sender
- Checking reorgs
- bumping fees

ekrembal:
# Description

Creates a child transaction for CPFP and submits the tx with submitpackage

## Linked Issues

- Closes #465 
- Closes #520


mmtftr:
TapSighashTypes were all set to Default in #511 which fixed a signature issue in that PR. You might want to wait for that to merge or do the same changes to test if that fixes the signing issue..

ekrembal:
> TapSighashTypes were all set to Default in #511 which fixed a signature issue in that PR. You might want to wait for that to merge or do the same changes to test if that fixes the signing issue..

Signing issue is fixed with Default sighash, but let's merge that first

mmtftr (commented):
Can we move the services (bitcoin_syncer, tx_sender) into `services/` and actors (operator, verifier, watchtower, actor) into `actors/`

core/src/bitcoin_syncer.rs
```diff
@@ -0,0 +1,205 @@
...
+    mode: BitcoinSyncerPollingMode,
+) -> Result<ChainHeadPollingResult, BridgeError> {
+    let (tx, _) = broadcast::channel(100);
+
+    let returned_tx = tx.clone();
+
+    let mut block_height = db
+        .get_max_height(None)
+        .await?
+        .ok_or(BridgeError::BlockNotFound)?;
```
mmtftr, line 117: this doesn't handle the first start does it? What happens when we first start the syncer with zero block entries in the db?

core/src/bitcoin_syncer.rs
```diff
@@ -0,0 +1,205 @@
...
+
+                let block_info = BlockInfo {
+                    block_hash,
+                    block_header,
+                    block_height,
+                };
+                new_blocks.push(block_info);
+
+                block_height -= 1;
+            }
```
mmtftr, line 152: this logic is too complicated. we need to simplify it. It also depends on a global variable (block_height).

I'm also not too sure on its correctness here. `block_height` starts off as the longest chain in our db. Then for each block that you can't find, you'll subtract the height by one going back on the longest chain. I feel like the `block_height` ought to start from block_height + 1 (which is the height of the first block_header that we fetch).

Overall, this logic is way too complicated and uses global variables which really complicates things.

I believe this can be extracted to something like:

```rust
async fn get_next_block(db) -> (Vec<Blocks> /* reorg */, Option<Block> /* one new block */)
```

core/src/bitcoin_syncer.rs
```diff
@@ -0,0 +1,205 @@
...
+    let returned_tx = tx.clone();
+
+    let mut block_height = db
+        .get_max_height(None)
+        .await?
+        .ok_or(BridgeError::BlockNotFound)?;
+
+    let handle = tokio::spawn(async move {
+        loop {
+            let mut block_hash = rpc.client.get_block_hash(block_height + 1).await?;
```
mmtftr, line 121: Is `?` right here? Wouldn't this return early if there were no block? Instead, I believe we want to wait for poll_delay then go back to the beginning of the loop.

core/src/bitcoin_syncer.rs
```diff
@@ -0,0 +1,205 @@
...
+                    block_header,
+                    block_height,
+                };
+                new_blocks.push(block_info);
+
+                block_height -= 1;
+            }
+
+            // If we haven't found a match after 100 blocks, the database is too far out of sync
+            if new_blocks.len() == 100 {
```
mmtftr, line 155: duplicated check, let's just make the above `for` a `loop` then break using this `if`

core/src/bitcoin_syncer.rs
```diff
@@ -0,0 +1,205 @@
...
+pub enum BitcoinSyncerEvent {
+    NewBlocks(Vec<BlockInfo>),
+    NewBlocksWithTxs(Vec<BlockInfoWithTxs>),
+    ReorgedBlocks(Vec<(BlockHash, BlockHash)>),
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum BitcoinSyncerPollingMode {
+    SyncOnly,
+    SyncAndPollTxs,
```
mmtftr, line 42: I feel like it should be
```
pub enum BitcoinSyncerMode {
  SyncHeaders,
  SyncHeadersAndTxs
}
```
PollingMode doesn't sound right here.

core/src/bitcoin_syncer.rs
mmtftr: Actually, I guess the second item would be a `Vec<Block>` of new blocks. It might be good to have an enum return:

```rust
enum BlockPollResult {
  OneNewBlock(BlockHeader),
  ReorgedBlocks(Vec<Block>, Vec<Block>)
}
```

core/src/database/bitcoin_syncer.rs
```diff
@@ -0,0 +1,131 @@
+use super::{wrapper::BlockHashDB, Database, DatabaseTransaction};
+use crate::{errors::BridgeError, execute_query_with_tx};
+use bitcoin::BlockHash;
+use std::{ops::DerefMut, str::FromStr};
+
+impl Database {
+    pub async fn set_chain_head(
```
mmtftr, line 7: `insert_chain_head`

core/src/database/bitcoin_syncer.rs
```diff
@@ -0,0 +1,131 @@
...
+    ) -> Result<Option<u64>, BridgeError> {
+        let query =
+            sqlx::query_as("SELECT height FROM bitcoin_syncer ORDER BY height DESC LIMIT 1");
+        let result: Option<(i64,)> =
+            execute_query_with_tx!(self.connection, tx, query, fetch_optional)?;
+        Ok(result.map(|(height,)| height as u64))
+    }
+
+    /// Gets the height from the block hash.
+    pub async fn get_height_from_block_hash(
```
mmtftr, line 38: `get_block_height`?

core/src/database/bitcoin_syncer.rs
```diff
@@ -0,0 +1,131 @@
...
+        let query = sqlx::query_as("SELECT height FROM tx_sender_block_info WHERE block_hash = $1")
+            .bind(block_hash.to_string());
+
+        let height: Option<(i64,)> =
+            execute_query_with_tx!(self.connection, tx, query, fetch_optional)?;
+        Ok(height.map(|(h,)| h as u64))
+    }
+
+    /// Gets the block hashes that have height bigger then the given height and deletes them.
+    pub async fn delete_chain_head_from_height(
```
mmtftr, line 52: `delete_chain_heads_after_height`?

core/src/bitcoin_syncer.rs
mmtftr: also better naming for function:
```rs
async fn poll_blocks()
```

core/src/bitcoin_syncer.rs
ceyhunsen: Looks like it is fixed with https://github.com/chainwayxyz/clementine/pull/509/commits/96409e2f1e876619df0dab0a5cd173f4e4e43c63

core/src/tx_sender.rs
```diff
@@ -0,0 +1,622 @@
...
+        let tx = builder.get_cached_tx().clone();
+        Ok(tx)
+    }
+    #[tokio::test]
+    #[serial_test::serial]
+    async fn test_create_fee_payer_tx() {
+        let (tx_sender, rpc, db, signer, network) = create_test_tx_sender().await;
+
+        let _bitcoin_syncer_handle =
+            bitcoin_syncer::start_bitcoin_syncer(db, rpc.clone(), Duration::from_secs(1))
```
ceyhunsen, line 567: We are in a test. We shouldn't waste time but waste CPU cycles with a lower sleep value.

ceyhunsen (changes requested):
Looks like tx sender is dependent on bitcoin syncer. But adding both of them in a single PR make them both harder to review. We should've been introduced bitcoin syncer first and than tx sender afterwards IMO.

core/src/tx_sender.rs
ceyhunsen: This is repeated multiple times in tests.

core/src/bitcoin_syncer.rs
```diff
@@ -0,0 +1,251 @@
...
+    pub block_hash: BlockHash,
+    pub block_header: Header,
+    pub block_height: u64,
+}
+
+/// Events emitted by the Bitcoin syncer.
+#[derive(Clone, Debug)]
+pub enum BitcoinSyncerEvent {
+    NewBlock(BlockHash),
+    ReorgedBlock(BlockHash),
```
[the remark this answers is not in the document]
ceyhunsen: It is discussed that this behavior is required in a private conversation.

PR #510, opened 2025-02-07
[WIP] add chain head polling background job

doscortados:


core/src/background.rs
```diff
@@ -0,0 +1,60 @@
...
+        .await?;
+    tx.commit().await?;
+    Ok(())
+}
+
+impl Database {
+    pub async fn set_tx_sender_chain_head(
+        &self,
+        tx: &mut sqlx::Transaction<'_, sqlx::Postgres>,
+        block_hash: String,
```
ekrembal, line 38: blockhash should be a type of bitcoin::BlockHash, and you can move this in db/tx_sender.rs

ekrembal (changes requested):

core/src/background.rs
doscortados: done

PR #511, opened 2025-02-07
[Actor, TxHandler] add signing of txhandler through tagged signatures

mmtftr:
# Description

Implements `partial_sign` and `partial_sign_commit` that signs the inputs that it supports.
## Linked Issues

- Closes #TBD

## Testing

Added signing for actor key/script spend scripts.

## Docs

Added some rustdoc on functions that explain the behavior, but still missing some docs.


core/src/builder/script.rs
```diff
@@ -162,8 +189,11 @@ impl SpendableScript for TimelockScript {
 }
 
 impl TimelockScript {
-    fn generate_witness(&self, signature: schnorr::Signature) -> Witness {
-        Witness::from_slice(&[signature.serialize()])
+    pub fn generate_script_inputs(&self, signature: &Option<schnorr::Signature>) -> Witness {
```
kpp, line 192: Better use `Option<&schnorr::Signature>`

core/src/builder/transaction/operator_collateral.rs
```diff
@@ -184,11 +192,15 @@ pub fn create_kickoff_timeout_txhandler(
 ) -> Result<TxHandler, BridgeError> {
     let builder = TxHandlerBuilder::new()
         .add_input(
+            NormalSignatureKind::KickoffTimeout1,
             kickoff_tx_handler.get_spendable_output(3)?,
+            SpendPath::ScriptSpend(1),
```
kpp, line 197: What's that 1? Magic number?

core/src/builder/transaction/txhandler.rs
```diff
@@ -211,43 +270,6 @@ impl TxHandler<Unsigned> {
         Ok(())
     }
 
-    // Candidate refactoring
```
kpp, line 214: Do you need to keep this code? Are you really going to use it in the future?

core/src/builder/transaction/operator_collateral.rs
mmtftr: yeah, we have a script indexing problem. We refer to scripts by their index. We should be tagging them someway else, open to suggestions here.

core/src/builder/transaction/txhandler.rs
mmtftr: I was thinking of doing something like this instead of script indexing (i.e. using magic numbers). We didn't do that. I don't really think this is an optimal solution (as it requires looping over all scripts), so decided to just remove it.

Also don't understand the comment here, do you want it back? I removed the lines (did not add them)

mmtftr: Since we're now using SpendPaths, we'd need to embed some kind of script selector closure (like `Box<dyn FnMut(&dyn SpendableScript) -> bool>`) into the SpendPath. That's going to be kind of expensive and slow

ekrembal (approved):

ekrembal (approved):

ekrembal (approved):

PR #547, opened 2025-02-17
Tx Sender with prerequisites

### LINKED ISSUE #541: Tx Sender with prerequisites and cancellations

ekrembal:
Closes #541


core/src/rpc/aggregator.rs
```diff
@@ -397,11 +398,12 @@ impl Aggregator {
...
-        let _fee_payer_utxo = self
-            .tx_sender
-            .create_fee_payer_utxo(*move_txhandler.get_txid(), move_tx.weight())
+
+        let mut dbtx = self.db.begin_transaction().await?;
+        self.tx_sender
+            .try_to_send(&mut dbtx, move_tx, FeePayingType::CPFP, &[], &[], &[])
             .await?;
-        self.tx_sender.save_tx(move_tx).await?;
+        dbtx.commit().await.expect("Failed to commit transaction");
```
ceyhunsen, line 406: can be mapped to status

scripts/schema.sql
```diff
@@ -249,4 +226,67 @@ create table if not exists bitcoin_syncer_event_handlers (
     primary key (consumer_handle)
 );
 
+
+-------- TX SENDER --------
```
ceyhunsen, line 230: I was thinking we can move this SQL file to database module's folder and separate them based on the contexts. It shouldn't stay on the same folder as the testing scripts.

Not this PR's job obviously, just a small note.

core/src/database/tx_sender.rs
```diff
@@ -48,337 +194,377 @@ impl Database {
...
-        assert!(db
-            .get_fee_payer_tx(None, non_existent, script_pubkey)
-            .await
-            .unwrap()
-            .is_empty());
+        .bind(id);
+
+        let result = execute_query_with_tx!(self.connection, tx, query, fetch_one)?;
+        Ok((
+            deserialize(&result.0).expect("Failed to deserialize tx"),
```
ceyhunsen, line 453: can be mapped to bridge error

core/src/tx_sender.rs
```diff
@@ -430,10 +596,9 @@ impl TxSender {
                     self.db
                         .save_fee_payer_tx(
                             None,
-                            bumped_txid,
+                            bumped_id,
                             new_txid_result.expect("New txid result is None"),
```
ceyhunsen, line 600: can be _new_txid, no?

core/src/database/tx_sender.rs
```diff
@@ -48,337 +194,377 @@ impl Database {
...
+            result.1,
+        ))
     }
 }
+
+// #[cfg(test)]
```
ceyhunsen, line 459: There are a lot of changes to this module. Why comment tests, instead of modifying them?

ceyhunsen (changes requested):

scripts/schema.sql
ekrembal: Yeah not this PR's job, but i actually like every table in a single file

core/src/database/tx_sender.rs
ekrembal: Added tests

core/src/bitcoin_syncer.rs
```diff
@@ -24,8 +24,8 @@ struct BlockInfo {
 /// Events emitted by the Bitcoin syncer.
 #[derive(Clone, Debug)]
 pub enum BitcoinSyncerEvent {
-    NewBlock(BlockHash),
```
mmtftr, line 27: what is this u32? block id? would be better to document in rustdoc

core/src/database/bitcoin_syncer.rs
```diff
@@ -66,23 +83,19 @@ impl Database {
         &self,
         tx: Option<DatabaseTransaction<'_, '_>>,
         height: u64,
-    ) -> Result<Vec<BlockHash>, BridgeError> {
```
mmtftr, line 69: can we not use u32? we had u32 in the event?

core/src/database/bitcoin_syncer.rs
```diff
@@ -289,9 +220,9 @@ impl Database {
         .await?;
 
         // Step 3: Retrieve the next event that hasn't been processed yet
-        let event = sqlx::query_as::<_, (i32, BlockHashDB, String)>(
+        let event = sqlx::query_as::<_, (i32, i32, String)>(
```
mmtftr, line 223: can we not request u32 from the db?

mmtftr (changes requested):
Looks good to me overall, I think we just need to address @ceyhunsen 's comments on the unnecessary `except` calls that can be converted into recoverable errors

core/src/tx_sender.rs
```diff
@@ -33,6 +32,25 @@ pub struct TxSender {
...
 }
 
+#[derive(Copy, Clone, Debug, Eq, Hash, PartialEq, PartialOrd, Ord, sqlx::Type)]
+#[sqlx(type_name = "fee_paying_type", rename_all = "lowercase")]
+pub enum FeePayingType {
+    CPFP,
+    RBF,
+}
+
+#[derive(Copy, Clone, Debug, Eq, Hash, PartialEq, PartialOrd, Ord)]
```
mmtftr, line 42: needs docs

core/src/tx_sender.rs
```diff
@@ -33,6 +32,25 @@ pub struct TxSender {
...
+pub struct ActivedWithTxid {
+    pub txid: Txid,
+    pub timelock: bitcoin::Sequence,
+}
+
+#[derive(Copy, Clone, Debug, Eq, Hash, PartialEq, PartialOrd, Ord)]
+pub struct ActivedWithOutpoint {
```
mmtftr, line 49: needs docs

core/src/database/tx_sender.rs
```diff
@@ -2,83 +2,212 @@
...
-    /// Saves a fee payer transaction to the database.
-    ///
-    /// # Arguments
-    /// * `bumped_txid` - The txid of the bumped transaction
-    /// * `fee_payer_txid` - The txid of the fee payer transaction
-    /// * `vout` - The output index of the UTXO
-    /// * `script_pubkey` - The script pubkey of the UTXO
-    /// * `amount` - The amount in satoshis
-    pub async fn save_fee_payer_tx(
+    pub async fn confirm_transactions(
```
mmtftr, line 18: this query looks really unoptimized and very complicated. Please write clear docs on what each query does (beyond the SQL operation like update/delete).

We should also look into understanding this operation's performance, optimizing this query, investigating required indices.

core/src/database/tx_sender.rs
```diff
@@ -2,83 +2,212 @@
...
+        .bind(i32::try_from(block_id).map_err(|e| BridgeError::ConversionError(e.to_string()))?)
+        .execute(tx.deref_mut())
+        .await?;
 
         Ok(())
     }
 
-    /// Updates the confirmation status of a fee payer transaction.
-    pub async fn confirm_fee_payer_tx(
+    pub async fn unconfirm_transactions(
```
mmtftr, line 137: same as before, please explain the operations

mmtftr (approved):
overall we need strong documentation on tx sender's behavior, both as a module and on the individual functions. I'm ok with the changes if the behavior is as expecetd and tests pass

ceyhunsen (approved):

PR #552, opened 2025-02-19
[Test] Remove macros in test in favor of fns

mmtftr:
# Description

<details>
<summary>OLD description</summary>

> Adds a new feature called "test" that integration tests use. Adds 
> 
> TODO: 
> - [x] We need to add a binary target with compile_error! 
> - [ ] Document the new "test" feature requirement (including VS code config changes)
> - [ ] Add tests for binary targets that checks prevention of build with "test" feature.
</details>

Moves integration tests into unit tests (cargo integration test feature/configurations are not sufficient for us to prevent inclusion in binaries).


mmtftr:
integration tests automatically build binaries which makes it impossible to run them at the moment with this setup. We may want to extract the integration tests into a separate crate entirely.

mmtftr:
@ceyhunsen what do you think?

mmtftr:
I don't think it restricts things, since we already have a workspace with multiple crates, and `cargo test` can be flagged to run all crates at the same time. We'd also have an advantage doing this since we can set a default feature of `clementine_core/test` enabled inside the test crate

ceyhunsen (commented):
Do you think we can also move `test_utils` in to `itest/common`?

ceyhunsen (approved):

PR #563, opened 2025-02-25
[Config] Implement Protocol Paramsets

### LINKED ISSUE #556: Add a protocol_paramset config and define paramsets

mmtftr:
# Description

This PR introduces a new config param named `protocol_paramset` which defines the various constants used within our bitcoin bridge contract. These include transaction UTXO amounts, timelock durations, btc network, winternitz commitment params

## Linked Issues

- Closes #556 

## Testing

Will add tests that validate that txhandlers can be constructed for all paramsets.
> Thinking of adding tests to check the spendability of txhandlers by using a mechanism that checks whether a Tx has an anchor output, if so can it be spent using CPFP (TRUC)? Otherwise, check if all of its inputs are anyonecanpay, so that it can be RBF'd.

## Docs

Docs are available in the new `protocol.rs` file.


core/src/config/protocol.rs
```diff
@@ -0,0 +1,151 @@
...
+    /// Collateral funding amount for operators used to fund the round transaction chain.
+    pub collateral_funding_amount: Amount,
+    /// Length of the blockhash commitment in kickoff transactions.
+    pub kickoff_blockhash_commit_length: u32,
+    /// Length of the message used in watchtower challenge transactions.
+    pub watchtower_challenge_message_length: usize,
+    /// Winternitz derivation log_d (shared for all WOTS commitments)
+    ///
+    /// Currently used in statics and thus cannot be different from [`WINTERNITZ_LOG_D`].
+    pub winternitz_log_d: u32,
```
atacann, line 68: Not used anymore

core/src/builder/transaction/challenge.rs
```diff
@@ -110,7 +111,7 @@ pub fn create_operator_challenge_nack_txhandler(
                 ),
                 kickoff_txhandler.get_spendable_output(challenge_ack_vout)?,
                 SpendPath::ScriptSpend(0),
-                Sequence::from_height(BLOCKS_PER_WEEK * 3),
+                Sequence::from_height(paramset.operator_challenge_nack_timelock),
```
atacann, line 114: change to paramset.operator_challenge_nack_timelock

core/src/config/protocol.rs
```diff
@@ -0,0 +1,151 @@
...
+    /// Number of blocks for watchtower challenge NACK timelock (currently BLOCKS_PER_WEEK / 2)
+    pub watchtower_challenge_nack_timelock: u16,
```
atacann, line 70: not used

core/src/config/protocol.rs
mmtftr: still used in `utils.rs:COMBINED_ASSERT_DATA`

core/src/builder/transaction/challenge.rs
mmtftr: that's already what it is though? I think it may be fixed in one of the latest commits

core/src/builder/transaction/operator_reimburse.rs
```diff
@@ -128,7 +132,7 @@ pub fn create_kickoff_txhandler(
                     .expect("taptree with one node at depth 1 will accept a script node")
                     .add_leaf(
                         1,
-                        TimelockScript::new(Some(nofn_xonly_pk), BLOCKS_PER_WEEK * 4)
+                        TimelockScript::new(Some(nofn_xonly_pk), paramset.assert_timeout_timelock)
```
atacann, line 135: nofn_4week.to_script_buf()

core/src/config/protocol.rs
mmtftr: removed

core/src/builder/transaction/operator_reimburse.rs
mmtftr: done

ceyhunsen (changes requested):
Looks like `ProtocolParamset` can derive `#[derive(Copy)]`. We can just pass that struct everywhere without passing as a reference. We can also get rid of `'static` lifetimes with this.

core/src/config/protocol.rs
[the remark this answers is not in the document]
mmtftr:
```rs
---- config::protocol::tests::test_size stdout ----
[core/src/config/protocol.rs:155:9] std::mem::size_of::<ProtocolParamset>() = 88
```
No, the `&'static ProtocolParamset` is an optimization. We only pass around a static reference (usize length, mostly 8 bytes).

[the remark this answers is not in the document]
mmtftr: second ask is done

ceyhunsen (approved):
I wouldn't say it was really that important to pass reference instead of copying but it is fine

core/src/config/protocol.rs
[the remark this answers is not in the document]
ceyhunsen: do we pass this struct in a critical loop that will slow down the operation?

PR #550, opened 2025-02-18
Check Citrea for withdrawals

ceyhunsen:
# Description

Describe what this pull request does, here.

## Linked Issues

- Closes #531 

## Testing

Describe how these changes were tested. If you've added new features, have you
added unit tests?

## Docs

Describe where this code is documented. If it changes a documented interface,
have the docs been updated?

PR #570, opened 2025-02-26
Make deposit and withdrawal at Citrea

### LINKED ISSUE #531: Implement a job to sync withdrawal_utxo from Citrea
### Issue Description

When a user makes a withdrawal in Citrea, it emits an event Withdrawal, we need to have a job in Citrea, where it syncs all these events, gets the move_txids, writes them into a db. https://github.com/chainwayxyz/citrea/blob/nightly/crates/evm/src/evm/system_contracts/src/Bridge.sol#L148

Later, we will use this when sending our kickoff txs.

### Tasks

_No response_

### Deadline of the Issue

-

ceyhunsen:
# Description

- Adds Citrea deposit test functions
- Adds Citrea withdrawal test functions

Re-open of #550

## Tests

- Moves out Bitcoind creation from `create_actors` because we need to use citrea-e2e's Bitcoind in new tests

## Linked Issues

- Related #531


ekrembal (approved):

core/src/test/common/citrea/bitcoin_merke.rs
```diff
@@ -0,0 +1,174 @@
+use serde::{Deserialize, Serialize};
```
atacann, line 1: typo in file name

ekrembal: merke is a good name :)

ekrembal: reverse of ekrem

PR #571, opened 2025-02-26
[Test] Use Unix sockets for reliability

### LINKED ISSUE #548: [Test] Use Unix domain sockets for speed and reliability
Using tcp sockets in tests are slow and due to limitations we only have 64k ports available.

We should instead use unix sockets to increase speed and reliability of our tests.

mmtftr:
# Description

Some tests are flaky due to the many ports we allocate for each test with actors. TCP has an overhead and is quite a lot slower when bound by bandwidth. More importantly, due to the increasingly parallelized tests, we have flaky tests due to port collisions when allocating ports.

This PR migrates the random TCP port allocation to random Unix socket allocation by generalizing connection primitives and refactoring test utils

## Linked Issues

- Closes #548 

## Testing

The tests must execute without flakiness

## Docs

Docs have been updated for test utils.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1010,6 +1010,7 @@ dependencies = [
  "hex",
  "http 1.2.0",
  "hyper 1.6.0",
+ "hyper-util",
  "jsonrpsee",
  "lazy_static",
  "prost",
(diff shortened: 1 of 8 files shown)


mmtftr:
It seems that unix sockets are slower in our case. Not sure why, but this PR is still a must-have since it fixes flaky tests.

ekrembal (approved):

Issue #559, opened 2025-02-21
Tx Sender improvments


ekrembal:
### To Do

- [x] Function to get the status of a `try_to_send` tx  @mmtftr 
- [ ] Tx sender log explaining why a tx is not getting sent  
- [x] Documentation on module behavior  @mmtftr 
  - Sending of fee-payer UTXOs & possible failure cases  
  - Sending of txs & possible failure cases  
  - Reasons for this behavior  
- [ ] Unit test for `bump_fee_with_fee_rate`  @mmtftr 
- [ ] Fix bump fee for RBF’ed txs  
- [ ] Define how to send Watchtower challenge tx @ekrembal 
- [ ] Alerting when a tx is active for 20 blocks but tx_sender couldn't send it @ekrembal 
- [ ] Support for multiple fee payer utxos (currently only one is signed in create_child_tx)

PR #572, opened 2025-02-28
Round managment + Update bitvm dependency

### LINKED ISSUE #553: Implement finalized payout_tx handler in operator
fn handle_finalized_payout(dbtx: DatabaseTransaction, deposit_outpoint: Outpoint)

This should find a signed kickoff connector and handle all try_to_sends for that.

ekrembal:
# Description

Add 2 internal grpc endpoints for operator: finalized payout tx and end round
Add tx names and details in tx_sender for better logging
Handled tx_sender errors, when it panics it stops the whole process
Updated BitVM dependency to not use patches

## Linked Issues

- Closes #553 
- Related to #559 

## Testing

Tested that the operator can get a reimbursment after calling end-round


core/src/config/protocol.rs
```diff
@@ -109,28 +109,28 @@ pub const REGTEST_PARAMSET: ProtocolParamset = ProtocolParamset {
...
-    disprove_timeout_timelock: BLOCKS_PER_WEEK * 5,
-    assert_timeout_timelock: BLOCKS_PER_WEEK * 4,
-    operator_reimburse_timelock: BLOCKS_PER_DAY * 2,
-    watchtower_challenge_timeout_timelock: BLOCKS_PER_WEEK * 2,
+    operator_challenge_timeout_timelock: 100,
+    operator_challenge_nack_timelock: 300,
+    disprove_timeout_timelock: 250,
+    assert_timeout_timelock: 200,
+    operator_reimburse_timelock: 48,
+    watchtower_challenge_timeout_timelock: 200,
```
mmtftr, line 125: revert these to use constants

core/src/utils.rs
```diff
@@ -236,104 +112,443 @@ impl BitvmCache {
...
+    pub fn get_assert_scripts(
+        &self,
+        xonly_public_key: XOnlyPublicKey,
+    ) -> Vec<std::sync::Arc<dyn SpendableScript>> {
+        let mut scripts = Vec::new();
+        let first_script: Arc<dyn SpendableScript> = Arc::new(WinternitzCommit::new(
+            vec![
+                (self.latest_blockhash_pk.to_vec(), 20),
+                (self.challenge_sending_watchtowers_pk.to_vec(), 20),
+                (self.bitvm_pks.0[0].to_vec(), 32),
```
atacann, line 433: needs to be * 2 for all WinternitzCommits in this function, because it needs message (digit) length not byte length

core/src/utils.rs
```diff
@@ -236,104 +112,443 @@ impl BitvmCache {
...
     BitvmCache {
-        intermediate_variables,
         disprove_scripts: scripts,
         replacement_places,
     }
 }
 
-pub fn replace_disprove_scripts(winternitz_pk: &[Vec<[u8; 20]>]) -> Vec<ScriptBuf> {
+#[derive(Debug, Clone, BorshSerialize, BorshDeserialize, PartialEq, Eq)]
+pub struct ClementineBitVMPublicKeys {
```
ceyhunsen, line 236: I'd suggest moving this to a new module in `bitvm.rs`

core/src/utils.rs
ekrembal: Fixed in 6c51b02

core/src/utils.rs
```diff
@@ -471,3 +686,74 @@ pub fn initialize_logger(level: Option<LevelFilter>) -> Result<(), BridgeError>
 
     Ok(())
 }
+
+/// Monitors a JoinHandle and aborts the process if the task completes with an error.
+/// Returns a handle to the monitoring task that can be used to cancel it.
+pub fn monitor_task_with_abort<T: Send + 'static>(
+    task_handle: tokio::task::JoinHandle<Result<T, crate::errors::BridgeError>>,
+    task_name: &str,
+) -> tokio::task::JoinHandle<()> {
```
ceyhunsen, line 695: Returning this is not bad but I think it should return nothing. Therefore it must be infallible. If it returns something, caller will confuse itself to watch this handle too.

core/src/rpc/operator.rs
```diff
@@ -186,9 +189,52 @@ impl ClementineOperator for Operator {
...
+        let deposit_outpoint = request
+            .get_ref()
+            .deposit_outpoint
+            .clone()
+            .expect("Failed to get deposit outpoint");
+        let deposit_outpoint: OutPoint = deposit_outpoint
+            .try_into()
+            .expect("Failed to convert deposit outpoint to OutPoint");
+
+        let mut dbtx = self.db.begin_transaction().await?;
```
ceyhunsen, line 218: Why dbtx for a single operation? Problem of the funcstion signature

core/src/rpc/operator.rs
```diff
@@ -186,9 +189,52 @@ impl ClementineOperator for Operator {
...
+        Ok(Response::new(Empty {}))
+    }
+
+    #[tracing::instrument(skip(self), err(level = tracing::Level::ERROR), ret(level = tracing::Level::TRACE))]
+    async fn internal_end_round(
+        &self,
+        _request: Request<Empty>,
+    ) -> Result<Response<Empty>, Status> {
+        let mut dbtx = self.db.begin_transaction().await?;
+        self.end_round(&mut dbtx).await?;
```
ceyhunsen, line 236: Why dbtx for a single operation? Problem of the funcstion signature

core/src/operator.rs
```diff
@@ -981,6 +650,290 @@ impl Operator {
 
         Ok(hashes)
     }
+
+    pub async fn handle_finalized_payout<'a>(
+        &'a self,
+        dbtx: DatabaseTransaction<'a, '_>,
```
ceyhunsen, line 656: Can we make it optional, so RPC functions doesn't need to create a dbtx?

ceyhunsen (changes requested):

core/src/utils.rs
```diff
@@ -236,104 +112,443 @@ impl BitvmCache {
...
+        pk_type_idx: u8,
+    ) -> [[[u8; 20]; DIGIT_LEN]; PK_LEN] {
+        (0..PK_LEN as u16)
+            .map(|pk_idx| Self::get_dummy_wpk(pk_type_idx, pk_idx))
+            .collect::<Vec<_>>()
+            .try_into()
+            .expect("Should be able to convert to array")
+    }
+
+    pub fn create_dummy() -> Self {
```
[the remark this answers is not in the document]
ceyhunsen: also other stuff

core/src/rpc/operator.rs
ekrembal: This rpc is temporary, this function MUST take a db tx

core/src/rpc/operator.rs
ekrembal: This rpc is temporary, this function MUST take a db tx.

core/src/operator.rs
ekrembal: No, it should be concrete that this funciton operates in a db tx.

core/src/builder/script.rs
```diff
@@ -639,62 +638,62 @@ mod tests {
...
+    // #[tokio::test]
+    // async fn test_winternitz_commit_spendable() {
+    //     let mut config = create_test_config_with_thread_name(None).await;
+    //     let regtest = create_regtest_rpc(&mut config).await;
+    //     let rpc = regtest.rpc();
+
+    //     let kp = bitcoin::secp256k1::Keypair::new(&SECP, &mut rand::thread_rng());
+    //     let xonly_pk = kp.public_key().x_only_public_key().0;
+
+    //     let derivation = WinternitzDerivationPath::BitvmAssert(
```
atacann, line 650: Uncomment test, something like below works.
One other idea is that we can only store pk_type and idx here, have a function that returns byte_size of a pk_type inside the new bitvm struct, then we can get message lengths from pk_type so that we do not pass message length everywhere and its easy to change
```suggestion
let derivation = WinternitzDerivationPath::BitvmAssert(
            64,
            3,
            0,
            Txid::all_zeros(),
            ProtocolParamsetName::Regtest.into(),
       );
```

ekrembal: re-added the test in 6781fae

i think it is okay to send the message length for now?

core/src/utils.rs
ekrembal: 6781fae

ceyhunsen (approved):

PR #579, opened 2025-03-04
Enable `main.rs`

### LINKED ISSUE #308: Convert library to binary
### Proposal Description

Currently, Clementine is qualified as a library, by Rust directives (have a `lib.rs` and not a `main.rs`). But structurally, it is a binary. If it is expected other binaries to depend code on this library, then it is fine. But if it's not, there is a problem for developers: Because every module is declared as public and Rust allows unused functions in libraries, codebase has uncatched unused functions. Maybe other linting hints are also hidden.

If it is not wanted Clementine to be a library, `lib.rs` should be removed and `main.rs` should be created. Helper binaries can still be present in `bin/` directory.

ceyhunsen:
# Description

Moves `bin/server.rs` to `main.rs` and deletes `lib.rs`. Also moves code inside of `lib.rs` to `utils.rs`.

Clementine binary now can only spawn a single server per process.

Closes #308


core/src/main.rs
```diff
@@ -0,0 +1,106 @@
...
+    if args.watchtower_server {
+        handles.push(
+            create_watchtower_grpc_server(config)
+                .await
+                .expect("Can't create watchtower server")
+                .1,
+        );
+
+        println!("Watchtower server is started.");
+    }
```
mmtftr, line 100: Add error logs that verify invariants that we require: if any server is running, we need to run a verifier because that's where the tx sender runs. If there's nothing selected, print an error (handles.is_empty())

core/src/main.rs
```diff
@@ -0,0 +1,106 @@
...
+    let mut handles = vec![];
+
+    if args.verifier_server {
+        handles.push(
+            create_verifier_grpc_server(config.clone())
+                .await
+                .expect("Can't create verifier server")
+                .1,
+        );
+        config.port += 1;
```
mmtftr, line 62: can we please properly specify ports in the config for each server?

mmtftr (changes requested):

core/src/main.rs
ceyhunsen: @ekrembal future work or here?

ekrembal: here

core/src/main.rs
ceyhunsen: There is an error message for no flags (line 40).

We can also force spawning a verifier server even though flag is not specified. So it will be better to convert flags from servers to entities: A higher level interface. WDYGT? @ekrembal @mmtftr

ceyhunsen: Apparently we can start an actor (not verifier) without starting verifier in that same binary but start that binary before. So, no need to force user to specify `--verifier-server`.

ceyhunsen: No, they aren't. Yes an entity won't function properly without one. But all the entites should be aware that they have to run verifier. This structure is just a preference of @ekrembal but can be changed later on. I believe possibility to run servers on different processes help testing but not sure how.

ekrembal:
We will need lib.rs in cli

PR #580, opened 2025-03-04
state machines

atacann:
## Docs

have the docs been updated?


core/build.rs
```diff
@@ -50,6 +50,10 @@ fn compile_protobuf() {
     tonic_build::configure()
         .build_server(true)
         .build_client(true)
+        .type_attribute(
```
ceyhunsen, line 53: Does this effects auto generated code?

atacann: Yes, its for adding serialize/deserialize to the proto struct

atacann: Yes, its for adding serialize/deserialize to the proto struct

core/src/states/mod.rs
```diff
@@ -0,0 +1,556 @@
...
+        handler: T,
+        paramset: &'static ProtocolParamset,
+    ) -> Result<Self, BridgeError> {
+        let queue = PGMQueueExt::new_with_pool(db.get_pool()).await;
+        queue
+            .init()
+            .await
+            .map_err(|e| BridgeError::Error(format!("Error initializing pqmq queue: {:?}", e)))?;
+
+        queue.create(&Self::queue_name()).await.map_err(|e| {
```
ceyhunsen, line 113: Looks like PgmqError is used multiple times. Maybe we can auto convert this in to BridgeError?

mmtftr: honestly, I think I'll be converting everything into eyre::Result. We'll be able to do it progressively since BridgeError now contains an eyre::Report variant. After all refactors, we should be able to use eyre::Report/Result directly. If we ever need to match specific errors, eyre supports `downcast_ref` which should be enough for any such usecase

core/src/operator.rs
```diff
@@ -941,6 +993,121 @@ impl Operator {
...
+                self.send_asserts(kickoff_id, deposit_data, watchtower_challenges)
+                    .await?;
+            }
+            Duty::VerifierDisprove {
+                kickoff_id,
+                deposit_data,
+                operator_asserts,
+                operator_acks,
+            } => {
+                tracing::info!("Operator {} called verifier disprove with kickoff_id: {:?}, deposit_data: {:?}, operator_asserts: {:?}, operator_acks: {:?}", self.idx, kickoff_id, deposit_data, operator_asserts.len(), operator_acks.len());
```
ceyhunsen, line 1070: maybe we can add `todo!()` here so that a test will fail if it tries to use this functionality

core/src/test/full_flow.rs
```diff
@@ -576,20 +598,21 @@ pub async fn run_happy_path_2(config: &mut BridgeConfig, rpc: ExtendedRpc) -> Re
...
-    //     .unwrap();
-    // send_tx(
-    //     &tx_sender,
-    //     &tx_sender_db,
-    //     &rpc,
-    //     challenge_tx.raw_tx.as_slice(),
-    // )
-    // .await
-    // .context("failed to send challenge transaction")?;
+    //Add later when RBF is implemented
```
ceyhunsen, line 601: comment can be deleted

core/src/tx_sender.rs
```diff
@@ -1068,10 +1073,14 @@ mod tests {
...
         let (tx_sender, rpc, db, signer, network) = create_test_tx_sender(rpc).await;
 
-        let _bitcoin_syncer_handle =
-            bitcoin_syncer::start_bitcoin_syncer(db.clone(), rpc.clone(), Duration::from_secs(1))
-                .await
-                .unwrap();
+        let _bitcoin_syncer_handle = bitcoin_syncer::start_bitcoin_syncer(
+            db.clone(),
+            rpc.clone(),
+            Duration::from_secs(1),
```
ceyhunsen, line 1079: note for the future: we can lower this for tests; no need to save computing power in tests

core/src/operator.rs
ceyhunsen: these log only match legs are wip, right? there more to them

mmtftr: there are issues for these atm. they'll be done by Oguz/Ozan I believe. todo! will break the tests so not doing this

core/src/test/full_flow.rs
mmtftr: done (to be pushed soon)

PR #596, opened 2025-03-10
Add handle_finalized

### LINKED ISSUE #531: Implement a job to sync withdrawal_utxo from Citrea

### LINKED ISSUE #594: Implement handle_finalized_block for Owner trait
We will both implement this for verifier and operator,
Operator's one will do nothing,
Verifier's one will update the db with Citrea syncs and Payout tx syncs

ekrembal:
Closes #594 
Closes #531


core/src/verifier.rs
```diff
@@ -1165,4 +1191,121 @@ impl Owner for Verifier {
...
     }
+
+    async fn handle_finalized_block(
+        &self,
+        mut dbtx: DatabaseTransaction<'_, '_>,
+        block_id: u32,
+        block_height: u32,
+        block: &bitcoin::Block,
+    ) -> Result<(), BridgeError> {
+        if let Some(citrea_client) = self.citrea_client.as_ref() {
```
mmtftr, line 1202: use guarded if-let for less nesting
```rs
let Some(citrea_client) = ... else { 
  return Err()
};
```

core/src/verifier.rs
```diff
@@ -1165,4 +1191,121 @@ impl Owner for Verifier {
...
+                self.db
+                    .set_move_to_vault_txid_from_citrea_deposit(
+                        Some(&mut dbtx),
+                        idx as u32,
+                        &move_to_vault_txid,
+                    )
+                    .await?;
+            }
+            for (idx, withdrawal_utxo_outpoint) in new_withdrawals {
+                tracing::error!("Setting withdrawal utxo: {:?}", withdrawal_utxo_outpoint);
```
mmtftr, line 1255: please don't use tracing::error to surface your logs. Instead you can always filter logs from a specific module when running code. Example filter to enable all logs for this module:

```sh
export RUST_LOG=warn,clementine_core::verifier=trace
```

core/src/verifier.rs
```diff
@@ -1165,4 +1191,121 @@ impl Owner for Verifier {
...
+
+                payout_txs_and_payer_operator_idx.push((idx, payout_txid, operator_idx));
+            }
+
+            self.db
+                .set_payout_txs_and_payer_operator_idx(
+                    Some(&mut dbtx),
+                    payout_txs_and_payer_operator_idx,
+                )
+                .await?;
```
mmtftr, line 1302: nit: This function is a little long, I'd prefer the inner logic to be split. I believe we can have

```rs
async fn get_citrea_l2_height_range(citrea_client, l1_block_height) -> (u32, u32);
async fn update_citrea_withdrawals(citrea, l2_height_range);
async fn update_finalized_payouts(...);
```

scripts/schema.sql
```diff
@@ -205,6 +205,20 @@ create table if not exists tx_sender_activate_try_to_send_outpoints (
...
+-------- FINALIZED BLOCK SYNCER , CITREA DEPOSITS AND WITHDRAWALS --------
+create table if not exists withdrawals (
+    idx int primary key,
+    move_to_vault_txid text not null check (move_to_vault_txid ~ '^[a-fA-F0-9]{64}'),
+    withdrawal_utxo_txid text not null check (withdrawal_utxo_txid ~ '^[a-fA-F0-9]{64}'),
+    withdrawal_utxo_vout int not null,
+    withdrawal_batch_proof_bitcoin_block_height int,
+    payout_txid text check (payout_txid ~ '^[a-fA-F0-9]{64}'),
+    payout_payer_operator_idx int,
+    is_payout_handled_by_operator boolean not null default false,
```
mmtftr, line 218: nit: `is_payout_handled`

core/src/verifier.rs
```diff
@@ -121,6 +124,28 @@ impl Verifier {
         )
         .await?;
 
+        let citrea_client = if !config.citrea_rpc_url.is_empty()
```
mmtftr, line 127: why not `CitreaClient::from_config` here? this logic doesn't belong in the verifier file

core/src/states/task.rs
```diff
@@ -198,9 +199,9 @@ impl<T: Owner + std::fmt::Debug + 'static> StateManager<T> {
                     Err(e) => {
                         tracing::error!("State manager run loop error: {:?}", e);
                         num_consecutive_errors += 1;
-                        if num_consecutive_errors > 50 {
-                            return Err(eyre::eyre!("Too many consecutive state machine errors"));
-                        }
+                        // if num_consecutive_errors > 50 {
+                        //     return Err(eyre::eyre!("Too many consecutive state machine errors"));
+                        // }
```
mmtftr, line 204: ?

mmtftr (changes requested):

ceyhunsen (changes requested):

core/src/verifier.rs
ekrembal: I didn't know this, i will get rid of tracing:errors no worries. Thanks

core/src/test/deposit_and_withdraw_e2e.rs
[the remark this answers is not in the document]
ekrembal: We can delete the deposit.rs and withdrawal.rs tests now, since deposit_and_withdarwal test does both. But let's keep them for now. since they also have additional asserts to check balances etc.

core/src/verifier.rs
ekrembal: Let's do this in #603

core/src/test/deposit_and_withdraw_e2e.rs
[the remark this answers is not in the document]
ceyhunsen: What do you think about moving these 2 files contents to this file?

ceyhunsen (approved):

core/src/database/verifier.rs
```diff
@@ -37,4 +40,245 @@ impl Database {
 
         Ok(pks.into_iter().map(|(_, pk)| pk.0).collect())
     }
+
+    pub async fn set_move_to_vault_txid_from_citrea_deposit(
+        &self,
+        tx: Option<DatabaseTransaction<'_, '_>>,
+        idx: u32,
```
mmtftr, line 47: what idx is this? it'd be better to name it. if it's a unique database id, then id would be better. if it's tied to citrea, citrea_idx would be better

core/src/database/verifier.rs
```diff
@@ -37,4 +40,245 @@ impl Database {
...
+        .bind(TxidDB(*move_to_vault_txid));
+
+        execute_query_with_tx!(self.connection, tx, query, execute)?;
+        Ok(())
+    }
+
+    pub async fn set_withdrawal_utxo_from_citrea_withdrawal(
+        &self,
+        tx: Option<DatabaseTransaction<'_, '_>>,
+        idx: u32,
```
mmtftr, line 66: same

core/src/test/common/mod.rs
```diff
@@ -20,6 +20,67 @@ use tonic::Request;
...
+/// Wait for a transaction to be in the mempool and than mines a block to make
+/// sure that it is included in the next block.
+///
+/// # Parameters
+///
+/// - `rpc`: The RPC client to use.
+/// - `txid`: The txid to wait for.
+/// - `tx_name`: The name of the transaction to wait for.
+/// - `timeout`: The timeout in seconds.
+pub async fn wait_tx_to_be_in_mempool_and_mine_block(
```
mmtftr, line 32: commit_tx_to_blockchain?

core/src/database/verifier.rs
```diff
@@ -37,4 +40,245 @@ impl Database {
...
+                    vout: u32::try_from(vout)?,
+                })
+            })
+            .transpose()
+    }
+
+    /// Returns the withdrawal indexes and their spending txid for the given
+    /// block id.
+    pub async fn get_payout_txs_from_citrea_withdrawal(
+        &self,
```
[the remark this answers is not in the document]
mmtftr: these are not finalized right?

core/src/verifier.rs
```diff
@@ -1077,6 +1101,142 @@ impl Verifier {
...
+        let payout_txids = self
+            .db
+            .get_payout_txs_from_citrea_withdrawal(Some(dbtx), block_id)
+            .await?;
+
+        let txid_to_tx_hashmap: HashMap<bitcoin::Txid, &bitcoin::Transaction> = block
+            .txdata
+            .iter()
+            .map(|tx| (tx.compute_txid(), tx))
+            .collect();
```
mmtftr, line 1212: I wonder if reusing BlockCache makes sense here?

core/src/verifier.rs
```diff
@@ -1173,4 +1333,46 @@ impl Owner for Verifier {
...
+        if self.citrea_client.is_none() {
+            return Ok(());
+        }
+
+        let citrea_client = self
+            .citrea_client
+            .as_ref()
+            .ok_or_else(|| BridgeError::Error("Citrea client is not available".to_string()))?;
+
+        let max_attempts = light_client_proof_wait_interval_secs.unwrap_or(600);
```
mmtftr, line 1354: 600 should go into a `const` definition somewhere

core/src/verifier.rs
```diff
@@ -1077,6 +1101,142 @@ impl Verifier {
         }
         Ok(())
     }
+
+    async fn get_citrea_l2_height_range(
```
mmtftr, line 1105: some docs explaining that this waits for max_attempts seconds at most

core/src/test/common/mod.rs
ekrembal: I don't think commit_tx_to_blockchain is a good name

core/src/verifier.rs
ekrembal: it might makes sense, but how i can pass it to here

core/src/database/verifier.rs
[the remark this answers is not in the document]
ekrembal: get_payout_txs_for_withdrawal_utxos

core/src/verifier.rs
ekrembal: Fixed

core/src/verifier.rs
ekrembal: Fixed

core/src/verifier.rs
mmtftr: no need to pass, just create one

ekrembal: but i don't need that much detailed info

PR #597, opened 2025-03-10
bridge circuit refactor tests and docs

aoengin:


circuits-lib/src/bridge_circuit/groth16.rs
```diff
@@ -1,94 +1,154 @@
...
+        let a = ark_bn254::G1Affine::deserialize_compressed(a_compressed)?;
+        let b = ark_bn254::G2Affine::deserialize_compressed(b_compressed)?;
+        let c = ark_bn254::G1Affine::deserialize_compressed(c_compressed)?;
+
+        Ok(CircuitGroth16Proof::new(a, b, c))
+    }
 
-    let mut claim_digest = Sha256::digest(concatenated);
-    claim_digest.reverse();
+    pub fn to_compressed(&self) -> Result<[u8; 128], SerializationError> {
```
mmtftr, line 60: Can this actually fail? If we're validating all elements when creating, why do we need to return a result here? If this cannot fail, let's replace unwraps with `except` and explain why it can't fail. If this can fail, then let's return the error instead of unwrapping

circuits-lib/src/bridge_circuit/groth16_verifier.rs
```diff
@@ -0,0 +1,95 @@
...
+    PREPARED_VK,
+};
+use super::groth16::CircuitGroth16Proof;
+use hex::ToHex;
+use sha2::{Digest, Sha256};
+use std::str::FromStr;
+
+pub fn create_output_digest(total_work: &[u8; 16]) -> [u8; 32] {
+    let total_work_digest: [u8; 32] = Sha256::digest(total_work).into();
+    let len_output: [u8; 2] = hex::decode("0200").unwrap().try_into().unwrap();
```
mmtftr, line 19: can we write the length in a u16 then encode? 

If len == 2, then it'd be:
```rs
let len_output: u16 = 2;

 let output_pre_digest: [u8; 98] = [
        &OUTPUT_TAG,
        &total_work_digest[..],
        &ASSUMPTIONS[..],
        &len_output.to_le_bytes(),
];
```

circuits-lib/src/bridge_circuit/groth16_verifier.rs
```diff
@@ -0,0 +1,95 @@
...
+    .try_into()
+    .expect("slice has correct length");
+
+    Sha256::digest(output_pre_digest).into()
+}
+
+pub fn create_claim_digest(output_digest: &[u8; 32], pre_state: &[u8; 32]) -> [u8; 32] {
+    let data: [u8; 8] = [0; 8];
+
+    let claim_len: [u8; 2] = [4, 0];
```
mmtftr, line 37: same as above

circuits-lib/src/bridge_circuit/storage_proof.rs
```diff
@@ -73,9 +112,73 @@ fn storage_verify(storage_proof: &EIP1186StorageProof, expected_root_hash: [u8;
...
+    use super::*;
+
+    const STORAGE_PROOF: &[u8] =
+        include_bytes!("../../../bridge-circuit-host/bin-files/storage_proof.bin");
+
+    #[test]
+    fn test_verify_storage_proofs() {
+        let storage_proof: StorageProof = borsh::from_slice(STORAGE_PROOF).unwrap();
+
+        let state_root: [u8; 32] = [
```
mmtftr, line 135: nit: `hex::decode()`

mmtftr (changes requested):
Thanks a lot, I have some minor comments. Also as an overall comment: can you change `anyhow` to `eyre`? They're almost API-compatible so should be a simple find-replace. You can check eyre docs for how to migrate some API differences.

ozankaymak (approved):

bridge-circuit-host/src/bridge_circuit_host.rs
```diff
@@ -127,33 +165,135 @@ pub fn prove_bridge_circuit(
...
+///
+/// This function will panic if:
+/// - Decoding `payout_tx` or `payment_block` fails.
+/// - Any block header chunk is not 80 bytes long.
+/// - Generating the Merkle or MMR proof fails.
+///
+pub fn create_spv(
+    payout_tx: &mut &[u8],
+    headers: &[u8],
+    payment_block: &[u8],
```
ekrembal, line 201: Why not use bitcoin::Block here?

bridge-circuit-host/src/bridge_circuit_host.rs
```diff
@@ -127,33 +165,135 @@ pub fn prove_bridge_circuit(
...
+
+    SPV {
+        transaction: payout_tx.into(),
+        block_inclusion_proof: payout_tx_proof,
+        block_header: block.header.into(),
+        mmr_inclusion_proof: mmr_inclusion_proof.1,
+    }
+}
+
+/// Generates a Groth16 proof for work-only verification.
```
ekrembal, line 239: Generates a Groth16 proof of a bitcoin header chain proof where it only outputs the total work?

And prove_work_only is not a good name, maybe we can use:
prove_work_only_header_chain_proof?

bridge-circuit-host/src/bridge_circuit_host.rs
```diff
@@ -127,33 +165,135 @@ pub fn prove_bridge_circuit(
...
+        .build()
+        .unwrap();
     let prover = default_prover();
     prover
         .prove_with_opts(env, WORK_ONLY_ELF, &ProverOpts::groth16())
         .unwrap()
         .receipt
 }
 
+/// Prepares the public inputs for the bridge circuit.
```
ekrembal, line 268: Good doc, but the name can change

bridge-circuit-host/src/bridge_circuit_host.rs
```diff
@@ -204,25 +344,28 @@ fn public_inputs(input: BridgeCircuitInput) -> SuccinctBridgeCircuitPublicInputs
     }
 }
 
+pub fn generate_winternitz(
```
ekrembal, line 347: sign witnernitz?

Also do we need this function here?
Can this be moved to tests?

circuits-lib/src/bridge_circuit/mod.rs
```diff
@@ -32,76 +30,51 @@ pub const IS_TEST: bool = {
...
+/// 11. Ensures the withdrawal transaction ID matches the input reference in `payout_spv.transaction`.
+/// 12. Computes the `deposit_constant` using the last output of the payout transaction.
+/// 13. Extracts and truncates the latest block hash and the payout transaction’s block hash.
+/// 14. Computes a Blake3 hash over concatenated block hash and watchtower flags.
+/// 15. Generates a final journal hash using Blake3 over concatenated data and commits it.
+///
+/// # Panics
+///
+/// - If the method ID in `hcp` does not match `HEADER_CHAIN_METHOD_ID`.
+/// - If `total_work` is greater than `hcp.chain_state.total_work`.
```
ekrembal, line 62: if mac total_work given by watchtowers is greater than hcp.chain_state.total_work?

circuits-lib/src/bridge_circuit/mod.rs
```diff
@@ -32,76 +30,51 @@ pub const IS_TEST: bool = {
...
+/// - If the SPV proof is invalid.
+/// - If the storage proof verification fails.
+/// - If the withdrawal transaction ID does not match the referenced input in `payout_spv`.
```
ekrembal, line 65: Don't we have a check for operator_id?

ekrembal (changes requested):

ekrembal:
Overall really good work!

PR #606, opened 2025-03-13
Decrease verifier count and calculate BRIDGE_PARAMS dynamically

### LINKED ISSUE #602: Use 4 verifier in tests

ceyhunsen:
# Description

Adds a `lazy_static` that calculates n-of-n dynamically and decreases verifier count to 4. Also, updates the docker config to latest.

Closes #602


scripts/docker/docker_config.toml
```diff
@@ -36,46 +41,47 @@ operator_wallet_addresses = [
...
-    "5555555555555555555555555555555555555555555555555555555555555555",
-    "6666666666666666666666666666666666666666666666666666666666666666",
-    "7777777777777777777777777777777777777777777777777777777777777777",
 ]
 all_operators_secret_keys = [
     "1111111111111111111111111111111111111111111111111111111111111111",
     "2222222222222222222222222222222222222222222222222222222222222222",
     "3333333333333333333333333333333333333333333333333333333333333333",
 ]
-all_watchtowers_secret_keys = [
```
atacann, line 74: Dont delete this

atacann (commented):
LGTM, there are some parameters in the toml configs that do not exist anymore in BridgeConfig like timeout_block_count, but I guess the parse test doesn't fail if toml has extra variables. Maybe we can check for that too?

ceyhunsen:
> I guess the parse test doesn't fail if toml has extra variables

Yes, that is correct. This behavior is derived from the external library, we don't do anything special. If you think we should fail while reading config file if there is an extra parameter, please open an issue.

scripts/docker/docker_config.toml
ceyhunsen: There is a problem with this change: #609

atacann (approved):

PR #612, opened 2025-03-14
Test BitVM Additional Disprove Script

aoengin:


core/src/test/bitvm_script.rs
```diff
@@ -0,0 +1,305 @@
...
+            &TEST_GROTH16_PUBLIC_INPUT.to_vec(),
+        );
+
+        let payout_tx_block_hash_witness = WINTERNITZ_MESSAGE_VERIFIER.sign(
+            &payout_tx_block_hash_params,
+            &payout_tx_block_hash_wsk,
+            &BRIDGE_CIRCUIT_BITVM_TEST_INPUTS
+                .payout_tx_block_hash
+                .to_vec(),
+        );
```
mmtftr, line 270: extract the common test code to a utility function

aoengin: I'll postpone the requested change until all tests are finalized, as the common parts may change after Ozan's or Ekrem's review.

ozankaymak (commented):
LGTM

core/src/test/bitvm_script.rs
ozankaymak: Agreed, reducing duplicate code would be good.

core/src/bitvm_client.rs
```diff
@@ -469,16 +474,21 @@ impl ClementineBitVMPublicKeys {
...
+                WinternitzDerivationPath::BitvmAssert(16 * 2, 4, 361, txid, paramset),
+                WinternitzDerivationPath::BitvmAssert(16 * 2, 4, 362, txid, paramset),
             ]
-        } else if (1..=3).contains(&mini_assert_idx) {
+        } else if (1..=2).contains(&mini_assert_idx) {
             // for 1, we will have 5 derivations index starting from 0 to 4
             // for 2, we will have 5 derivations index starting from 5 to 9
             // for 3, we will have 5 derivations index starting from 10 to 13
-            let derivations: u32 = (mini_assert_idx as u32 - 1) * 5;
+            let derivations: u32 = (mini_assert_idx as u32 - 1) * 6;
```
ekrembal, line 487: Can you change the comments above

ekrembal (changes requested):

core/src/citrea/mod.rs
```diff
@@ -472,6 +472,7 @@ mod tests {
     }
 
     #[tokio::test]
+    #[ignore = "This test is not working as expected"]
```
ceyhunsen, line 475: It works as expected but is too flaky. Maybe change it to something like:

```
Includes code that won't change much and the test itself is too flaky; Ignoring...
```

bridge-circuit-host/src/bridge_circuit_host.rs
```diff
@@ -500,6 +500,12 @@ mod tests {
 
         let (ark_groth16_proof, output_scalar_bytes_trimmed, bridge_circuit_bitvm_inputs) =
             prove_bridge_circuit(bridge_circuit_host_params, TEST_BRIDGE_CIRCUIT_ELF);
+        println!("ARK_GROTH16_PROOF: {:?}", ark_groth16_proof);
+        println!("OUTPUT_SCALAR_BYTES: {:?}", output_scalar_bytes_trimmed);
+        println!(
+            "BRIDGE_CIRCUIT_BITVM_INPUTS: {:?}",
+            bridge_circuit_bitvm_inputs
+        );
```
mmtftr, line 508: are we supposed to merge these?

core/src/test/full_flow.rs
```diff
@@ -840,6 +840,7 @@ mod tests {
     use super::*;
 
     #[tokio::test(flavor = "multi_thread")]
+    #[serial_test::serial]
```
mmtftr, line 843: add module documentation that explains why we added this

mmtftr (commented):
Looks good to me, minor comments

core/src/test/full_flow.rs
ceyhunsen: Maybe module documentation will be missed by some. I rather prefer function documentations

mmtftr: i dont like duplicating the same explanation. can we do module docs that explain and function comments that point to the module docs like

// see top documentation for why serial is used

ceyhunsen: If the comment will be long, sure. If not, isn't this also duplicating the comments? Either way, I am OK. Just wanted to point out many people won't look up for module documentation about why there is a `serial_test::serial` present for these tests

aoengin: There should not be there anymore, thanks for reminding me to remove.

mmtftr (approved):
Looks good to me,  thanks

ekrembal (approved):

PR #614, opened 2025-03-14
Replacement deposit

atacann:
Replacement deposits will be used when we want to update bridge design. From the old move tx, we create a new deposit tx, and sign all the kickoff transactions again with the new deposit as usual. Old deposit will be unusable as old movetx output is spent.
Also adds docs to duties.


core/src/rpc/clementine.proto
```diff
@@ -83,8 +83,24 @@ message WinternitzPubkey {
   repeated bytes digit_pubkey = 3;
 }
 
-// A deposit request's details.
 message DepositParams {
+  oneof deposit_data {
+    OriginalDeposit original_deposit = 1;
```
mmtftr, line 88: I don't think OriginalDeposit is an appropriate name here, Original can be confused to mean "unique" when used on its own. Something like BaseDeposit may be better.

mmtftr (changes requested):
I'd appreciate an explanation of the behavioral changes in the PR description.

ekrembal (approved):

Issue #532, opened 2025-02-14
Implement a job to sync finalized payout tx


ekrembal:
Operators will only take action when there is a finalized payout tx,
every operator will have a config variable named payout_tx_finalization_depth,
this job will actively search withdrawal_utxos from #531 If it sees a finalized payout tx with the given operator_id, it will start the process of sending kickoff_tx and others

ekrembal:
Closed with #616

Issue #603, opened 2025-03-12
Implement MockCitreaClient for testing without Citrea


ekrembal:
https://github.com/chainwayxyz/clementine/pull/596#discussion_r1988495412

ekrembal:
Closed with #616

PR #616, opened 2025-03-17
Devnet fixes

ekrembal:
Closes: #532 #603

Opened #627 for the requested change


core/src/test/common/mod.rs
```diff
@@ -34,7 +34,7 @@ pub async fn mine_once_after_in_mempool(
     txid: Txid,
     tx_name: Option<&str>,
     timeout: Option<u64>,
-) -> Result<(), BridgeError> {
+) -> Result<usize, BridgeError> {
```
ceyhunsen, line 37: This new return type should be explained in function's documentation

core/src/test/deposit_and_withdraw_e2e.rs
```diff
@@ -293,3 +375,59 @@ async fn citrea_deposit_and_withdraw_e2e() -> Result<()> {
     );
     TestCaseRunner::new(CitreaDepositAndWithdrawE2E).run().await
 }
+#[tokio::test]
```
ceyhunsen, line 378:
```suggestion
#[tokio::test]
#[ignore = "Manual testing utility"]
```

ceyhunsen (approved):

core/src/test/deposit_and_withdraw_e2e.rs
```diff
@@ -293,3 +375,59 @@ async fn citrea_deposit_and_withdraw_e2e() -> Result<()> {
...
+        .send_to_address(&deposit_address.0, config.protocol_paramset().bridge_amount)
+        .await?;
+
+    // wait until the deposit tx is in a block
+    rpc.mine_blocks(1).await.unwrap();
+
+    println!("Deposit address: {:?}", deposit_address);
+
+    // gRPC request:
+    println!("grpcurl -plaintext -proto core/src/rpc/clementine.proto -d '{{");
```
mmtftr, line 422: what is this?

core/src/config/protocol.rs
```diff
@@ -24,57 +27,24 @@ pub enum ProtocolParamsetName {
     Mainnet,
     Regtest,
     Testnet4,
+    Signet,
 }
 
 impl From<ProtocolParamsetName> for &'static ProtocolParamset {
     fn from(name: ProtocolParamsetName) -> Self {
+        if std::env::var("PROTOCOL_CONFIG_PATH").is_ok() {
```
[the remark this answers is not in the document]
ceyhunsen: Don't think this is needed. This won't block users from using this env var if they really want.

ceyhunsen (changes requested):
ci not passing and didnt notice before. it must be fixed

core/src/test/deposit_and_withdraw_e2e.rs
ceyhunsen: i think this is a manual deposit call string for testing deposit endpoint outside of rust tests

Issue #481, opened 2025-01-28
Error Handling Tracker


mmtftr:
### Proposal Description

We have problems with error handling. This issue is a tracker that includes sub-issues that aim to improve error handling and visibility across the codebase.

Some overarching concerns:
- We use `unwrap()`s for Error's that can happen in all likelihood.
  - Solution: we should only be using panicking error handler functions when we are sure that the error has no chance of happening. Eg. casting a Sha256 hash to `[u8; 32]`
- We only have one error type, and it's way too cluttered. It's also not used properly everywhere.
  - Solution: separate the Error types into subtypes that we can easily compose using `thiserror`'s `#[from]` attribute.
- We don't utilize `source`/`backtrace` features of `std::error`.
  - Solution: we should look into using libraries like `eyre` to build up error reports.
- We don't separate user errors from internal errors clearly, which can become a downstream problem for end users of the binaries.
  - Solution (#304): separate errors into Internal errors and user errors
  - Solution (extra): aid the user in solving the issues when it's a user error, generate an error debugging report when it's an internal error.

ozankaymak:
https://docs.rs/eyre/latest/eyre/

PR #625, opened 2025-03-18
refactor(error): replace #[from] variants with wrap_err

### LINKED ISSUE #624: Separate BridgeError into module errors
### Issue Description

This issue is to separate BridgeErrors into modules (although in the process, eyre::Report migration will also be partially done to avoid extra work in the future)

### Tasks

Parts of the task:
- [ ] Remove #from variants
- [ ] Move relevant variants into the module as a module-only error that is then returned inside an eyre::Report.

### Deadline of the Issue

-

mmtftr:
# Description

Modules should have independent error types that are composed with `eyre::Report`. BridgeError will be split into their respective modules' error types.

#### Tentative
BridgeError will have a variant for all errors and will be included inside eyre::Report's

## Linked Issues

- Closes #624
- Related to #481

## Docs

> TODO
Docs will be updated in the errors module. We include module docs for the general error handling paradigms (eyre, composition, conversion, wrap_err, module-specific errors), and each module's error type should include docs on the possible errors.

### DIFF
--- a/core/src/bitcoin_syncer.rs
+++ b/core/src/bitcoin_syncer.rs
@@ -3,6 +3,7 @@
 //! This module provides common utilities to fetch Bitcoin state. Other modules
 //! can use this module to operate over Bitcoin.
 
+use eyre::Context;
 use crate::{
     config::protocol::ProtocolParamset,
     database::{Database, DatabaseTransaction},
@@ -42,8 +43,8 @@ async fn fetch_block_info_from_height(
     rpc: &ExtendedRpc,
     height: u32,
 ) -> Result<BlockInfo, BridgeError> {
-    let hash = rpc.client.get_block_hash(height as u64).await?;
-    let header = rpc.client.get_block_header(&hash).await?;
+    let hash = rpc.client.get_block_hash(height as u64).await.wrap_err("Failed to get block hash")?;
+    let header = rpc.client.get_block_header(&hash).await.wrap_err("Failed to get block header")?;
 
     Ok(BlockInfo {
         hash,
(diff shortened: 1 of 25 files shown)


core/src/bitcoin_syncer.rs
```diff
@@ -42,8 +43,16 @@ async fn fetch_block_info_from_height(
     rpc: &ExtendedRpc,
     height: u32,
 ) -> Result<BlockInfo, BridgeError> {
-    let hash = rpc.client.get_block_hash(height as u64).await?;
-    let header = rpc.client.get_block_header(&hash).await?;
+    let hash = rpc
+        .client
+        .get_block_hash(height as u64)
+        .await
+        .wrap_err("Failed to get block hash")?;
```
ceyhunsen, line 50: Can't we replicate previous behavior? It is nicer to define an auto conversion (to a default message) for the common errors, instead of writing an error message for every single one like this one.

mmtftr: the previous setup did not always have specific errors like this. it'd have said "RPC error" instead of "Failed to get block hash".

We can automate this using custom error types in our own code with `thiserror::Error`'s Display impl but I don't think it's generally worth the effort. This also becomes self-documenting code (here, it's trivial when reading, but sometimes error messages in the code also explain the failure modes of the function).

It also enables us to search by error when we get an error in production. We may not have an exact backtrace in production so it'll be good to be able to search across the codebase for a unique error message. We can avoid `wrap_err` by returning eyre::Report (which will auto-convert) but I'd rather have the `wrap_err`s whenever possible.

mmtftr:
There are still BridgeError::Error's to migrate but I'm leaving those to another PR as this PR is getting too big. Ready for review (if CI passes)

core/src/bitcoin_syncer.rs
mmtftr: I've included instructions on how we'll be wrapping types with messages using `into_eyre` which also includes the extra messages we've written in the BridgeError variants.

ceyhunsen (approved):
Thanks for your efforts, mega PR

Issue #613, opened 2025-03-14
Close state manager for some tests


atacann:
State manager can break some tests like flow tests (Ex. state mngr auto sends challenge tx on happy_path_1 test). Write a way (rpc call -> backgroundmngr oneshot close channel?) to close the state managers then close them on relevant tests.

Alternatively change verifier, operator::new, create_actors etc so that we can choose what tasks(tx_sender, state machine, etc.) to start. This is probably better.

atacann:
Done with #630

PR #630, opened 2025-03-19
[Test] Citrea mock in-memory storage + State Machine killswitch

mmtftr:
# Description

- Implemented global in-memory storage for citrea in place of db usage.
- Removed the unused suffix argument for the test config util
- Implemented the option to switch off State Machines for certain tests https://github.com/chainwayxyz/clementine/issues/613


core/src/test/common/test_utils.rs
```diff
@@ -252,6 +252,7 @@ pub async fn create_test_config_with_thread_name(suffix: Option<&str>) -> Bridge
         };
 
     config.db_name = handle.to_string();
+    config.citrea_rpc_url = handle.to_string();
```
ceyhunsen, line 255: this is no no: should not include suffix. because all the servers need to access the same citrea server

ceyhunsen (changes requested):

core/src/test/common/test_utils.rs
mmtftr: the suffix is never used, I might as well remove it in this PR

ekrembal (approved):

core/src/config/mod.rs
```diff
@@ -20,6 +20,19 @@ use std::{fs::File, io::Read, path::PathBuf};
 
 pub mod protocol;
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct TestParams {
+    pub should_run_state_manager: bool,
```
ceyhunsen, line 25: We should move things like `all_verifiers_secret_keys` here too in a different pr, great stuff.

ceyhunsen (approved):

PR #641, opened 2025-03-21
Remove all public keys in config and return verifiers' public keys in `aggregator::setup`

### LINKED ISSUE #537: Remove all public keys in config, we should rely on pks distributed in aggregator setup
This should also handle that all operator's are set correctly

ceyhunsen:
# Description

- Removes public key fields in `BridgeConfig`.
- `aggregator::setup` returns verifiers' public keys

## Linked Issues

- Closes #537

### DIFF
--- a/core/src/aggregator.rs
+++ b/core/src/aggregator.rs
@@ -235,7 +235,7 @@ impl Aggregator {
     }
 
     #[tracing::instrument(skip(self), err(level = tracing::Level::ERROR), ret(level = tracing::Level::TRACE))]
-    fn aggregate_move_partial_sigs(
+    async fn aggregate_move_partial_sigs(
         &self,
         deposit_data: DepositData,
         agg_nonce: &MusigAggNonce,
@@ -251,16 +251,13 @@ impl Aggregator {
                 .to_byte_array(),
         );
 
-        let verifiers_public_keys = futures::executor::block_on(async {
-            Ok::<_, BridgeError>(
-                self.nofn
-                    .read()
-                    .await
-                    .clone()
-                    .ok_or(eyre!("N-of-N not set, yet"))?
-                    .public_keys,
-            )
-        })?;
+        let verifiers_public_keys = self
+            .nofn
+            .read()
+            .await
+            .clone()
+            .ok_or(eyre!("N-of-N not set, yet"))?
+            .public_keys;
 
         let final_sig = aggregate_partial_signatures(
             &verifiers_public_keys,
--- a/core/src/aggregator.rs
+++ b/core/src/aggregator.rs
@@ -19,6 +19,7 @@ use crate::{
 };
 use bitcoin::hashes::Hash;
 use bitcoin::secp256k1::{schnorr, Message};
+use eyre::eyre;
 use futures_util::future::try_join_all;
 use secp256k1::musig::{MusigAggNonce, MusigPartialSignature};
 use std::sync::Arc;
@@ -254,9 +255,9 @@ impl Aggregator {
         let verifiers_public_keys = self
             .nofn
             .try_read()
-            .map_err(|_| BridgeError::NofNNotSet)?
+            .map_err(|_| eyre!("Verifier index not set, yet"))?
             .clone()
-            .ok_or(BridgeError::NofNNotSet)?
+            .ok_or(eyre!("Verifier index not set, yet"))?
             .public_keys;
 
         let final_sig = aggregate_partial_signatures(
--- a/core/src/aggregator.rs
+++ b/core/src/aggregator.rs
@@ -256,7 +256,7 @@ impl Aggregator {
             .read()
             .await
             .clone()
-            .ok_or(BridgeError::Error(format!("N-of-N not set!")))?
+            .ok_or(BridgeError::Error("N-of-N not set!".to_string()))?
             .public_keys;
         let final_sig = aggregate_partial_signatures(
             &verifiers_public_keys,
--- a/core/src/aggregator.rs
+++ b/core/src/aggregator.rs
@@ -245,20 +245,22 @@ impl Aggregator {
             deposit_data,
             self.config.protocol_paramset(),
         )?;
-        // println!("MOVE_TX: {:?}", tx);
-        // println!("MOVE_TXID: {:?}", tx.tx.compute_txid());
+
         let message = Message::from_digest(
             tx.calculate_script_spend_sighash_indexed(0, 0, bitcoin::TapSighashType::Default)?
                 .to_byte_array(),
         );
 
-        let verifiers_public_keys = self
-            .nofn
-            .try_read()
-            .map_err(|_| eyre!("Verifier index not set, yet"))?
-            .clone()
-            .ok_or(eyre!("Verifier index not set, yet"))?
-            .public_keys;
+        let verifiers_public_keys = futures::executor::block_on(async {
+            Ok::<_, BridgeError>(
+                self.nofn
+                    .read()
+                    .await
+                    .clone()
+                    .ok_or(eyre!("N-of-N not set, yet"))?
+                    .public_keys,
+            )
+        })?;
 
         let final_sig = aggregate_partial_signatures(
             &verifiers_public_keys,
--- a/core/src/aggregator.rs
+++ b/core/src/aggregator.rs
@@ -1,13 +1,16 @@
+use std::sync::Arc;
+
 use crate::builder::transaction::DepositData;
 use crate::extended_rpc::ExtendedRpc;
 use crate::rpc::clementine::{DepositParams, OperatorKeysWithDeposit};
 use crate::tx_sender::TxSenderClient;
+use crate::verifier::NofN;
 use crate::{
     builder::{self},
     config::BridgeConfig,
     database::Database,
     errors::BridgeError,
-    musig2::{aggregate_partial_signatures, AggregateFromPublicKeys},
+    musig2::aggregate_partial_signatures,
     rpc::{
         self,
         clementine::{
@@ -17,10 +20,7 @@ use crate::{
     },
 };
 use bitcoin::hashes::Hash;
-use bitcoin::{
-    secp256k1::{schnorr, Message},
-    XOnlyPublicKey,
-};
+use bitcoin::secp256k1::{schnorr, Message};
 use futures_util::future::try_join_all;
 use secp256k1::musig::{MusigAggNonce, MusigPartialSignature};
 use tonic::Status;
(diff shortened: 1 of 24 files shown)


core/src/rpc/clementine.proto
```diff
@@ -340,7 +340,7 @@ service ClementineOperator {
 // Verifier --------------------------------------------------------------------
 
 message VerifierParams {
-  uint32 id = 1;
+  optional uint32 id = 1;
```
ceyhunsen, line 343: Verifier index is determined later down the flow. But we have to generate this struct just for the public key before assignation. So, I made it optional. I could also set this 0 or some other dummy value when returning verifier params. But it could become problematic if misused. Opinions are welcome.

ekrembal: Why verifier index is determined later down the flow?

ceyhunsen: Because at the time when `verifier::new` is called, a verifier doesn't know all the verifier public keys. After the `set_verifiers` is called, index is determined via the index of the pk in the list. Then we set index.

ekrembal: we have index in the config for that

ekrembal: I would say let's discuss this in the planning meeting

ceyhunsen: #651

core/src/rpc/aggregator.rs
```diff
@@ -528,8 +530,15 @@ impl Aggregator {
         )?;
 
         // aggregate partial signatures
+        let verifiers_public_keys = self
+            .nofn
+            .read()
```
mmtftr, line 535: This pattern of get and return error should be a struct method

core/src/rpc/aggregator.rs
```diff
@@ -705,7 +724,29 @@ impl ClementineAggregator for Aggregator {
...
+        let verifier_public_keys: Vec<PublicKey> = verifier_public_keys_clone
+            .clone()
+            .await?
+            .iter()
+            .map(|vpk| {
+                PublicKey::from_slice(vpk)
+                    .map_err(|e| Status::internal(format!("Failed to parse public key: {:?}", e)))
+            })
+            .collect::<Vec<Result<_, Status>>>()
+            .into_iter()
```
mmtftr, line 737: unnecessary double collect here, you should be able to collect to Result<Vec<_, _>> with the first call

core/src/rpc/aggregator.rs
```diff
@@ -705,7 +724,29 @@ impl ClementineAggregator for Aggregator {
...
+            .collect::<Result<Vec<_>, Status>>()?;
+        let signer = Actor::new(
+            self.config.secret_key,
+            self.config.winternitz_secret_key,
+            self.config.protocol_paramset().network,
+        );
+        let nofn = NofN::new(signer.public_key, verifier_public_keys.clone())?;
```
mmtftr, line 744: signer created only for the public key? refactor into config.calc_pubkey()

core/src/rpc/aggregator.rs
```diff
@@ -705,7 +724,29 @@ impl ClementineAggregator for Aggregator {
...
+        self.nofn.write().await.replace(nofn);
```
mmtftr, line 745: it'd be better to do the collection and setting of the nofn keys inside the `verifier_public_keys` future above. it's shared so that execution will be once and it'll be tidier.

mmtftr: at the same time, we should extract it into Aggregator::setup_verifier_keys

core/src/rpc/aggregator.rs
```diff
@@ -849,9 +890,16 @@ impl ClementineAggregator for Aggregator {
         ));
 
         // Start the signature aggregation pipe.
+        let verifiers_public_keys = self
+            .nofn
+            .read()
+            .await
+            .clone()
+            .ok_or(Status::internal("N-of-N not set"))?
```
mmtftr, line 898: common utility here

core/src/rpc/clementine.rs
```diff
@@ -1507,13 +1507,20 @@ pub mod clementine_aggregator_client {
...
+        /// 3. Collects all operator configs from each operator
+        /// 4. Distributes these operator configs to all verifiers
+        /// 5. Collects all possible Winternitz pubkeys (determined by operator idx, seqcol idx, kickoff idx) from each watchtower
+        /// 6. Distributes these Winternitz pubkeys to all verifiers
         pub async fn setup(
             &mut self,
             request: impl tonic::IntoRequest<super::Empty>,
-        ) -> std::result::Result<tonic::Response<super::Empty>, tonic::Status> {
+        ) -> std::result::Result<
+            tonic::Response<super::VerifierPublicKeys>,
```
mmtftr, line 1521: why return the pubkeys btw?

core/src/rpc/parser/verifier.rs
```diff
@@ -32,8 +32,17 @@ where
     type Error = Status;
 
     fn try_from(verifier: &Verifier<C>) -> Result<Self, Self::Error> {
+        let idx = verifier
+            .idx
+            .try_read()
+            .map_err(|_| Status::internal("Verifier index not set, yet"))?;
```
mmtftr, line 38: this looks like incorrect behavior to me, we should wait for the rwlock instead of trying once like this.

https://doc.rust-lang.org/std/sync/struct.RwLock.html#method.try_read

Also, for one-time initialization values, we can utilize [OnceLock](https://doc.rust-lang.org/std/sync/struct.OnceLock.html) which is made for this purpose and is guaranteed to be readable if there is a value.

I believe OnceLock is also applicable to the aggregator case, though dynamic verifier set is likely to make it a RwLock again.

mmtftr (changes requested):
Thanks, there are a couple of refactors we should do.

core/src/rpc/aggregator.rs
ceyhunsen: Yeah this was totally unnecessary: https://github.com/chainwayxyz/clementine/pull/641/commits/74a54d1f64d54a2db6f2f32821f34b2d72f34f41

core/src/rpc/aggregator.rs
ceyhunsen: https://github.com/chainwayxyz/clementine/pull/641/commits/74a54d1f64d54a2db6f2f32821f34b2d72f34f41

core/src/rpc/aggregator.rs
ceyhunsen: I can return converted pks in that future. But does it change anything? Because this loop also run once.

core/src/rpc/clementine.rs
mmtftr: I guess we can keep this. This also made me think about verifier initialization for the dynamic verifier set feature. We'll need an endpoint to return the current verifier set.

core/src/rpc/aggregator.rs
mmtftr: Yeah, the moving was to refactor it into a new function and so that related logic stays close together

ceyhunsen: https://github.com/chainwayxyz/clementine/pull/641/commits/f5aeeb9cd88fdb872657459bb889df7cac0c0499 wdyt?

core/src/rpc/parser/verifier.rs
ceyhunsen: https://github.com/chainwayxyz/clementine/pull/641/commits/fecb44db5d97ecaec4caf5228d790158f888907a

`blocking_read` was returning an error about something like blocking code in async execution context. This was the only thing that worked. But apparently this has it own problems. Fixed it.

core/src/rpc/aggregator.rs
ceyhunsen: But the type is `Arc<tokio::sync::RwLock<Option<NofN>>>`. Don't think we need a wrapper around this.

core/src/rpc/parser/verifier.rs
```diff
@@ -32,8 +32,15 @@ where
     type Error = Status;
 
     fn try_from(verifier: &Verifier<C>) -> Result<Self, Self::Error> {
+        let id = futures::executor::block_on(async {
+            match *verifier.idx.read().await {
+                Some(idx) => convert_int_to_another("id", idx, u32::try_from).map(Some),
+                None => Ok(None),
+            }
+        })?;
+
```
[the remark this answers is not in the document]
ceyhunsen: have you ever heard `TryFrom`

core/src/rpc/clementine.proto
```diff
@@ -463,10 +463,15 @@ message AggregatorWithdrawResponse {
...
-  // 1. Collecting verifier keys, then distributing all verifier keys to all verifiers
-  // 2. Collecting all operator configs, and distributing them to all verifiers
-  // 3. Collecting all possible Winternitz pubkeys (determined by operator idx, seqcol idx, kickoff idx) from watchtowers, and distributing to verifiers.
-  rpc Setup(Empty) returns (Empty) {}
+  //
+  // 1. Collects verifier keys from each verifier
+  // 2. Distributes these verifier keys to all verifiers
+  // 3. Collects all operator configs from each operator
+  // 4. Distributes these operator configs to all verifiers
+  // 5. Collects all possible Winternitz pubkeys (determined by operator idx, seqcol idx, kickoff idx) from each watchtower
```
ekrembal, line 471: This doc is still wrong, watchtower's don't have winternitz pks now

ceyhunsen: removed in https://github.com/chainwayxyz/clementine/pull/641/commits/8c9cacd6027cba033c2b6a6292a5f943085535ea

core/src/aggregator.rs
```diff
@@ -246,14 +245,25 @@ impl Aggregator {
...
         )?;
-        // println!("MOVE_TX: {:?}", tx);
-        // println!("MOVE_TXID: {:?}", tx.tx.compute_txid());
+
         let message = Message::from_digest(
             tx.calculate_script_spend_sighash_indexed(0, 0, bitcoin::TapSighashType::Default)?
                 .to_byte_array(),
         );
+
+        let verifiers_public_keys = futures::executor::block_on(async {
```
mmtftr, line 254: why are we doing this? this'll block a tokio thread and potentially cause problems.

mmtftr (changes requested):

core/src/aggregator.rs
ceyhunsen: because function is not async and to read `self.nofn`, i needed to call async functions. is there an alternative to this?

mmtftr: this is potentially a deadlock in single threaded runtimes and if runtime threads are exhausted. I'm not really ok merging this. We should make the function async. 

We can also make nofn a OnceLock as I commented earlier

ceyhunsen:
No approvals yet I merged it. My mindset is: Trust your code. I don't even test my code. You heard me right, I trust my code. CI/CD fails? Faulty computer I would say. This is how you develop software.

PR #643, opened 2025-03-24
Unspent kickoff sender

atacann:
- Adds automatically sending unspent kickoff tx's in case an operator sends ReadyToReimburse Tx without spending all of its kickoffs.
- As there was a problem with collecting wrong signatures of Unspent Kickoff Tx previously (one signature was getting signed with a tweaked key when it shouldn't), refactored signing so that it always needs the spend path, also added some data to SignatureInfo (returned from sighash fns) that holds merkle root hash if signature is a key spend sig. 
- Add a tweak cache to hold previously calculated key tweaks. This is created for new deposits because new deposits can run concurrently. It is its own struct instead of inside Actor because it is hard to mutably hold Actor without cloning.
- Change watchtower challenges to use keypath #645.


core/src/actor.rs
```diff
@@ -86,6 +90,94 @@ impl WinternitzDerivationPath {
...
+    ) -> Result<&Keypair, BridgeError> {
+        if let Some(saved_keypair) = self.saved_keypair {
+            if saved_keypair != *keypair {
+                return Err(BridgeError::Error(format!(
+                    "TweakCache used with a different keypair, saved: {:?}, given: {:?}",
+                    saved_keypair, keypair
+                )));
+            }
+        } else {
+            self.saved_keypair = Some(*keypair);
```
mmtftr, line 142: The first call saves the keypair, then the rest uses it? why not save it on creation?

core/src/actor.rs
```diff
@@ -86,6 +90,94 @@ impl WinternitzDerivationPath {
...
+        .add_tweak(
+            &SECP,
+            &TapTweakHash::from_key_and_tweak(pubkey, merkle_root).to_scalar(),
+        )?
+        .0)
+}
+
+#[derive(Debug, Clone, Default)]
+pub struct TweakCache {
+    // A cache that holds tweaked keys so that we do not need to repeatedly calculate them.
```
mmtftr, line 117: move the explanation to above the struct

mmtftr: also, I'm confused as to why this isn't a part of Actor, can we elaborate on that? If it's due to the cloning of it everywhere, we should also consider trying to integrate it into existing caches like the contractcontext etc.

core/src/builder/sighash.rs
```diff
@@ -78,12 +77,23 @@ pub struct PartialSignatureInfo {
     pub kickoff_utxo_idx: usize,
 }
 
+/// Contains information about the spend path of the utxo.
+/// If it is KeyPath, it also includes the merkle root hash of the scripts as
+/// the hash is needed to tweak the key before signing.
+#[derive(Copy, Clone, Debug)]
+pub enum TapTweakData {
```
mmtftr, line 84: I thought the purpose of this struct is to provide the information required to sign the UTXO (i.e. information required for a tweak or if a tweak is not required). It doesn't provide the spend path. Can we elaborate this in the docs?

core/src/actor.rs
atacann: Its mostly because to hold this in Actor you need actor to be mutable. To make actor mutable you need to clone it as mutconstantly or make almost all the functions in verifier, operator, tx sender etc &mut self, and clone owner/verifier as mut while passing to threads, I just thought this could be a different struct then and we can create it only when needed. ContractContext is valid for only one kickoff_id, but tweakcache is technically valid for all sign operations, we can create it once during deposit signing. We dont do this in musig2 partial sign tho

PR #650, opened 2025-03-25
refactor(test): create utility ensure_async

mmtftr:
# Description

Implemented a new utility to be used across testing code.

### DIFF
--- a/core/src/test/common/mod.rs
+++ b/core/src/test/common/mod.rs
@@ -38,11 +38,56 @@ pub mod citrea;
 mod setup_utils;
 pub mod tx_utils;
 
-pub async fn ensure_async(
+/// Polls a closure until it returns true, or the timeout is reached. Exits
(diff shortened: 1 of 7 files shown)


core/src/test/common/mod.rs
```diff
@@ -36,6 +38,29 @@ pub mod citrea;
 mod setup_utils;
 pub mod tx_utils;
 
+pub async fn ensure_async(
```
ceyhunsen, line 41: Don't have any suggestions but we should have a better name and function documentation for this; Name doesn't ring a bell.

ceyhunsen (approved):

core/src/test/common/mod.rs
mmtftr: renamed and added docs

ekrembal (approved):

PR #653, opened 2025-03-26
Remove `blockgazer`, run header chain prover in `verifier::handle_finalized_block` and prove in batches

### LINKED ISSUE #517: Remove code duplication in blockgazer.rs, use bitcoin_sycner.rs
Refactor header_chain_prover so that it uses broadcasted blocks

### LINKED ISSUE #644: Remove blockgazer module and move prover to verifier's `handle_finalized`
### Proposal Description

- [ ] Also remove first proof path in config

ceyhunsen:
# Description

Removes blockgazer and uses  `verifier::handle_finalized_block` to feed header chain prover and trigger proving. Also adds support for proving headers in batches.

## Linked Issues

- Closes #644 
- Closes #517


core/src/citrea/mod.rs
```diff
@@ -172,7 +172,7 @@ impl CitreaClient {
                 .wrap_err("Failed to get logs")?;
             logs.extend(logs_chunk);
 
-            from_height += to_height;
+            from_height = to_height + 1;
```
ceyhunsen, line 175: Irrelevant bug fix to this PR. Thanks for @atacann.

core/src/header_chain_prover.rs
```diff
@@ -0,0 +1,718 @@
...
+                current_block_header,
+                current_block_height,
+                previous_proof,
+            )
+            .await?;
+        tracing::info!(
+            "Receipt for block with hash {:?} and height with: {:?}: {:?}",
+            current_block_hash,
+            current_block_height,
+            receipt
```
mmtftr, line 363: how long is the to_string of this receipt in general (I believe the receipt contains our public outputs and the seal)? Will it fill up the logs?

core/src/header_chain_prover.rs
```diff
@@ -0,0 +1,718 @@
...
+    ///
+    /// - [`Receipt`]: Specified block's proof receipt
+    pub async fn get_header_chain_proof(&self, hash: BlockHash) -> Result<Receipt, BridgeError> {
+        match self.db.get_block_proof_by_hash(None, hash).await? {
+            Some(r) => Ok(r),
+            None => Err(HeaderChainProverError::NoHeaderChainProof(hash).into()),
+        }
+    }
+
+    /// Sets a new block to database, later to be proven.
```
mmtftr, line 292: "later to be proven" as in "the blockgazer will prove the saved block" right? also "save_new_block" is better for this

core/src/database/header_chain_prover.rs
```diff
@@ -201,110 +132,35 @@ mod tests {
...
-            db.get_latest_block_info(None).await.unwrap()
-        );
-
-        // Adding a new block with smaller height should not effect what's
-        // getting returned.
-        let smaller_height = height - 1;
-        block.header.time = 1; // To avoid same block hash.
-        db.set_new_block(None, block.block_hash(), block.header, smaller_height)
+        let height = 2;
+        db.set_new_block(None, block_hash, block.header, height)
```
mmtftr, line 156: might be out of scope but the naming of this function is very bad.

should be something like `db.set_blockgazer_block` or `db.save_blockgazer_block`. `set_new_block` is ambiguous in terms of who uses the block and for what.

core/src/header_chain_prover.rs
mmtftr: yeah sorry, I meant to say "the header chain prover will prove the saved block in its task"

core/src/database/header_chain_prover.rs
mmtftr: confused as before, `set_prover_block` etc.

core/src/header_chain_prover.rs
ceyhunsen: This will only be printed every 10 minutes for now and is not that long in my opinion. These log can be valuable in the future IMO.

Also this will be printed once every 100 blocks with the next PR, so even less of a problem.

Example output:

```
2025-04-07T08:38:37.646277Z  INFO clementine_core::header_chain_prover: core/src/header_chain_prover.rs:359: Receipt for block with hash 09ec3f19dd700cf0b8fdacde496d56195f2ea5f1f1fcc04b9a67e99d2c5c6842 and height with: 1: Receipt { inner: Fake(FakeReceipt { claim: MaybePruned { value: ReceiptClaim { pre: MaybePruned { value: SystemState { pc: 0x00207290, merkle_root: Digest(1b0a8a5d38a45b1d239836c41b4d8d782f33126a0091d7097604d105f9a6fbd2) }, digest: Digest(3b2c6d0835ba99a7b4eef5885f58213b85a032e7f69cecde90fda05eafefd1c8) }, post: MaybePruned { value: SystemState { pc: 0x00000000, merkle_root: Digest(0000000000000000000000000000000000000000000000000000000000000000) }, digest: Digest(a3acc27117418996340b84e5a90f3ef4c49d22c79e44aad822ec9c313e1eb8e2) }, exit_code: Halted(0), input: MaybePruned { digest: Digest(0000000000000000000000000000000000000000000000000000000000000000) }, output: MaybePruned { value: Some(Output { journal: MaybePruned { value: [59, 44, 109, 8, 53, 186, 153, 167, 180, 238, 245, 136, 95, 88, 33, 59, 133, 160, 50, 231, 246, 156, 236, 222, 144, 253, 160, 94, 175, 239, 209, 200, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 4, 66, 104, 92, 44, 157, 233, 103, 154, 75, 192, 252, 241, 241, 165, 46, 95, 25, 86, 109, 73, 222, 172, 253, 184, 240, 12, 112, 221, 25, 63, 236, 9, 255, 255, 127, 32, 0, 0, 0, 0, 218, 229, 73, 77, 10, 143, 243, 103, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 107, 187, 79, 91, 57, 137, 21, 41, 80, 188, 181, 210, 90, 165, 245, 98, 71, 40, 74, 28, 225, 69, 104, 200, 189, 139, 246, 123, 178, 180, 114, 113, 2, 0, 0, 0], digest: Digest(d222798484b3071e45686f517a143972cd6e753b5f5a944cbd29a996d06cc45d) }, assumptions: MaybePruned { value: Assumptions([]), digest: Digest(0000000000000000000000000000000000000000000000000000000000000000) } }), digest: Digest(647fab8ebb1281d0b20878dec7720e7d15099bdd0747a5f1ac3a42b5935c0744) } }, digest: Digest(7439a87c25dcf425b23590f0f17d7bc88b6a80aa2f64af22a6ccf91756466d80) } }), journal: Journal { bytes: [59, 44, 109, 8, 53, 186, 153, 167, 180, 238, 245, 136, 95, 88, 33, 59, 133, 160, 50, 231, 246, 156, 236, 222, 144, 253, 160, 94, 175, 239, 209, 200, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 4, 66, 104, 92, 44, 157, 233, 103, 154, 75, 192, 252, 241, 241, 165, 46, 95, 25, 86, 109, 73, 222, 172, 253, 184, 240, 12, 112, 221, 25, 63, 236, 9, 255, 255, 127, 32, 0, 0, 0, 0, 218, 229, 73, 77, 10, 143, 243, 103, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 107, 187, 79, 91, 57, 137, 21, 41, 80, 188, 181, 210, 90, 165, 245, 98, 71, 40, 74, 28, 225, 69, 104, 200, 189, 139, 246, 123, 178, 180, 114, 113, 2, 0, 0, 0] }, metadata: ReceiptMetadata { verifier_parameters: Digest(0000000000000000000000000000000000000000000000000000000000000000) } }
```

core/src/database/header_chain_prover.rs
ceyhunsen: So, the names were too general I admit. I made an effort to fix it with https://github.com/chainwayxyz/clementine/pull/653/commits/17b418f15347c6a4ef06a8094f1ab4d60551caee. Please tell me if it still can be confused with other stuff.

core/src/header_chain_prover.rs
ceyhunsen: I am not so sure if this should be marked as debug. Because some actors will directly use this data for their operations. So I think it will be useful for them to see if this data is ready and how it is.

core/src/database/header_chain_prover.rs
```diff
@@ -35,76 +34,30 @@ impl Database {
...
         let query = sqlx::query_as(
-            "SELECT height, block_hash FROM header_chain_proofs ORDER BY height DESC;",
+            "SELECT h1.block_hash,
+                    h1.block_header,
+                    h1.height,
+                    h2.proof
+                FROM header_chain_proofs h1
+                JOIN header_chain_proofs h2 ON h1.prev_block_hash = h2.block_hash
+                WHERE h2.proof IS NOT NULL
+                ORDER BY h1.height
```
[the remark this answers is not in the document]
ceyhunsen: good ai

core/src/header_chain_prover.rs
```diff
@@ -0,0 +1,787 @@
...
+            .await?
+            .ok_or(eyre::eyre!("No proven block found"))?;
+        let tip_height = self
+            .rpc
+            .client
+            .get_block_count()
+            .await
+            .wrap_err("Can't get block tip height")?;
+
+        if tip_height - latest_proven_block.2 > self.batch_size
```
atacann, line 297: tip_height - latest_proven_block.2 >= self.batch_size

core/src/header_chain_prover.rs
atacann: I think its better to log hash and height as info, receipt as debug or trace.

core/src/header_chain_prover.rs
```diff
@@ -0,0 +1,775 @@
...
+    /// prover, proves these blocks here.
+    ///
+    /// # Parameters
+    ///
+    /// - `hash`: Target block hash
+    ///
+    /// # Returns
+    ///
+    /// - [`Receipt`]: Specified block's proof receipt
+    pub async fn get_header_chain_proof(&self, hash: BlockHash) -> Result<Receipt, BridgeError> {
```
atacann, line 274: This function always proves until tip if block height is not a multiple of batch size. For example if 150th blocks proof is requested, and we have 100th and 200th blocks proof in db, it will still prove from 200th to current tip. 

Its possible we wont ever need any other proof except the tip but then the function should be changed to not require blockhash and change name to get_header_chain_proof_to_tip.

One other problem is that we should probably only prove finalized blocks, but asking tip height from rpc and getting headers from rpc will also include non finalized blocks. Instead of asking rpc you can get the current tip + block headers saved to db with save_unproven_block, as they will only include finalized blocks.

ceyhunsen: Great catch, thanks! Fixed it in https://github.com/chainwayxyz/clementine/pull/653/commits/89d8da4d44d930512139710e7687238f945193cd. I take the advice of to remove hash input and return confirmed tip every time. Please check for errors, again.

atacann: I still have the same concern. My advice is -> change _confirmed_ named functions to finalized, and do not call bitcoin rpc in get_tip_header_chain_proof at all. Only use blocks saved in db. (They are definitely finalized as they get saved to db in handle_finalized_block).

Using confirmed blocks instead of finalized ones have several problems:

- You prove latest confirmed block, set the proof in db. Then bitcoin reorgs but prover cant detect the reorg, but latest proved block still returns the reorged block. 
- I suspect you cant even save confirmed blocks proof in db, set_block_proof (or whatever it is) will fail as latest confirmed block will not even be in the db. It probably doesn't fail in test because testnet protocolparamset has finality depth of 1 (so confirmed = finalized), not 6.

ceyhunsen: It was a misnaming from my side. I was using only finalized blocks but using the name confirmed. Also RPC calls were only fetching finalized block headers and nothing else. But we have them on our db, so I used them instead of RPC calls and removed `ExtendedRpc` altogether from HCP.

https://github.com/chainwayxyz/clementine/pull/653/commits/c3bb7c1ce00da6cd02a0f8f8d2fab7cbec470105 wdyt?

atacann: 👍 I didn't see that you got tip height from db.

core/src/database/header_chain_prover.rs
```diff
@@ -35,76 +35,188 @@ impl Database {
...
+    /// # Returns
+    ///
+    /// Returns `None` if either no proved blocks are exists or blockchain tip
+    /// is already proven.
+    ///
+    /// - [`BlockHash`] - Hash of last block in the batch
+    /// - [`Header`] - Headers of the blocks
+    /// - [`u64`] - Height of the last block in the batch
+    /// - [`Receipt`] - Previous block's proof
+    pub async fn get_next_n_non_proven_block(
```
ekrembal, line 136: Why do we need this?

ekrembal (changes requested):
I think the logic should be just proving a batch and writing that proof to the DB, why we write non-proven blocks? There is too much code in database/header_chain_prover

core/src/header_chain_prover.rs
ekrembal, line 1: I think we only need the function prove_block_headers, we don't need rpc, we don't need to save unproven blocks

ekrembal (changes requested):

core/src/database/header_chain_prover.rs
ceyhunsen: To prepare a batch of block headers for proving in batches.

core/src/header_chain_prover.rs
ceyhunsen: How are we going to trigger proving when a batch is ready, than? Also, RPC is removed in the latest commit. We can use bitcoin_syncer's blocks but it will complicate either code or SQL.

core/src/database/header_chain_prover.rs
```diff
@@ -35,76 +35,188 @@ impl Database {
...
+    /// # Returns
+    ///
+    /// Returns `None` if either no proved blocks are exists or blockchain tip
+    /// is already proven.
+    ///
+    /// - [`BlockHash`] - Hash of the block
+    /// - [`Header`] - Header of the block
+    /// - [`u64`] - Height of the block
+    /// - [`Receipt`] - Previous block's proof
+    pub async fn get_next_non_proven_block(
```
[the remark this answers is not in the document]
ceyhunsen: https://github.com/chainwayxyz/clementine/pull/653/commits/3cfea97571f46f476d8f73adc6618b75e18f5450

core/src/database/header_chain_prover.rs
```diff
@@ -35,76 +35,188 @@ impl Database {
...
+    /// - [`Receipt`] - Previous block's proof
+    pub async fn get_next_n_non_proven_block(
+        &self,
+        count: u32,
+    ) -> Result<Option<(Vec<(BlockHash, Header, u64)>, Receipt)>, BridgeError> {
+        let next_non_proven_block = self.get_next_unproven_block(None).await?;
+        let next_non_proven_block = match next_non_proven_block {
+            Some(next_non_proven_block) => next_non_proven_block,
+            None => return Ok(None),
+        };
```
mmtftr, line 144: let Some(next_non_proven_block) = self.get_next_unproven_block(None).await? else {return Ok(None)};

is clearer

core/src/header_chain_prover.rs
```diff
@@ -0,0 +1,761 @@
...
+            // PS: This also ignores other db errors but there are other places
+            // where we check for those errors.
+            let _ = db
+                .save_unproven_finalized_block(
+                    None,
+                    block_hash,
+                    block_header,
+                    proof_output.chain_state.block_height.into(),
+                )
+                .await;
```
mmtftr, line 135: I'm against this ignoring. At the very least, we have to do an inspect_err with tracing::error

```suggestion
            let _ = db
                .save_unproven_finalized_block(
                    None,
                    block_hash,
                    block_header,
                    proof_output.chain_state.block_height.into(),
                )
                .await.inspect_err(|e| tracing::error!("Failed to save unproven finalized block to db {e:?}"));
```

core/src/header_chain_prover.rs
```diff
@@ -0,0 +1,761 @@
...
+
+    /// Proves blocks till the block with hash `current_block_hash`.
+    ///
+    /// # Parameters
+    ///
+    /// - `current_block_hash`: Hash of the block to prove
+    /// - `block_headers`: Block headers to prove
+    /// - `previous_proof`: Previous proof's receipt
+    #[tracing::instrument(skip_all)]
+    async fn prove_blocks(
```
mmtftr, line 160: Isn't this more like `prove_block_and_save()`. You're not proving or saving the previous blocks, they're just dependencies for the current_block.

core/src/verifier.rs
```diff
@@ -1365,6 +1370,11 @@ where
         self.update_finalized_payouts(&mut dbtx, block_id, &block_cache)
             .await?;
 
+        self.header_chain_prover
+            .save_unproven_block_cache(Some(&mut dbtx), &block_cache)
+            .await?;
+        self.header_chain_prover.prove_if_ready().await?;
```
mmtftr, line 1376: can this run in the background with tokio::spawn? @ekrembal @atacann what do you think? this'd potentially take a long time and stall the state machine. It might be better if it's done asynchronously.

mmtftr (commented):
Some docs comments and one possible improvement for the state manager.

core/src/verifier.rs
mmtftr: I feel like the header chain prover can be a background task with its own client like txsender. It's a better and clearer abstraction that way. `get` calls doing intensive computations reduces determinism in the codebase (any `get` call is potentially super expensive and there is a danger of race conditions: what if multiple places call `get` and prove at the same time?)

mmtftr: Now that I think about it, I don't think the `get` function should do proving. It's a possible DoS attack vector for us in the future. If a malicious actor can trigger many proof tasks at the same time for the same batch, they can overwhelm the CPU easily.

ceyhunsen: I agree with you and this was my previous implementation. But the feedbacks leaded to this design. We can keep `get` function to only return finalized tip's proof but other stuff can be reverted. @ekrembal @atacann your comments are welcome.

core/src/header_chain_prover.rs
```diff
@@ -0,0 +1,761 @@
...
+        tracing::debug!(
+            "Proof receipt for header chain circuit input {:?}: {:?}",
+            input,
+            receipt
+        );
+
+        Ok(receipt)
+    }
+
+    /// Get the proof of the latest finalized blockchain tip.
```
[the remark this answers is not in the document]
ceyhunsen: https://github.com/chainwayxyz/clementine/pull/653/commits/a6c21398ea2137ac7938d1c52ffbec099398d426

core/src/header_chain_prover.rs
ceyhunsen: https://github.com/chainwayxyz/clementine/pull/653/commits/a6c21398ea2137ac7938d1c52ffbec099398d426

core/src/header_chain_prover.rs
ceyhunsen: https://github.com/chainwayxyz/clementine/pull/653/commits/a6c21398ea2137ac7938d1c52ffbec099398d426

With the aggregator distributing initial header chain proof PR, this call will be removed. Because of this, I don't think it is worth to implement a proper error check here.

core/src/verifier.rs
atacann: Yes, something like tokio spawn blocking can also maybe prevent thread starvation on tests (not sure), but it needs careful implementation. For example if proving is in state machine, we definitely know if state machine passes the iteration we generated a watchtower challenge/operator assert. But if we move proving into some other part we need to make sure it is persistent (with a message queue again prob). I don't think we need to implement this if proving times are fast. We can look into it again after we check how long an operator's proof takes.

Also in general for watchtower challenges, we don't need to  constantly prove to tip but only get the latest proven block from db, given that its not out of sync (100 blocks ~ at most about 16 hours before). I don't think it impacts security too much. Operator needs to prove a specific height though.

ekrembal (approved):

aoengin (approved):

PR #654, opened 2025-03-28
[TxSender] Overhaul and refactor with RBF rewrite. Fee calculation, PSBT-based signing, debug logs

mmtftr:
# Description

This pull request aims to improve the Tx Sender module by adding better and more relevant logs, observability mechanisms for activated txs (txs that are supposed to be sent), and better testing.

## Linked Issues

- Related to #

## Testing

added unit tests?

## Docs

have the docs been updated?


mmtftr:
Moving to draft for a future release. The PR has grown too large for this deployment

core/src/database/tx_sender.rs
```diff
@@ -133,6 +161,39 @@ impl Database {
...
+        // Record debug info for confirmed RBF transactions
+        for (tx_id,) in confirmed_rbf_txs {
+            // Add debug state change
+
+            tracing::debug!(try_to_send_id=?tx_id,  "Transaction confirmed in block {}: RBF confirmation",
+            block_id);
+
+            // Update sending state
+            let _ = self
+                .update_tx_debug_sending_state(
```
ceyhunsen, line 189: Does this only here for debugging on a test env or we want this in a deployment, also?

core/src/database/tx_sender.rs
```diff
@@ -541,8 +621,164 @@ impl Database {
                 .map(u32::try_from)
                 .transpose()
                 .wrap_err("Failed to convert seen_block_id to u32")?,
+            serde_json::from_str(result.4.as_deref().unwrap_or("null")).wrap_err_with(|| {
+                format!("Failed to decode rbf_signing_info from {:?}", result.4)
+            })?,
         ))
     }
+
+    // Debug Functions
```
ceyhunsen, line 630: If we want to block these functions to be included in binaries, we should guard them with `#[cfg(test)]`

core/src/extended_rpc.rs
```diff
@@ -285,56 +385,49 @@ impl ExtendedRpc {
...
-        })
+        // Return the new txid
+        Ok(bump_fee_result
+            .txid
+            .ok_or_eyre("Failed to get Txid from bump_fee_result")?)
     }
 
+    /// Creates a new instance of the ExtendedRpc with a new client connection.
+    ///
+    /// This is useful when you need a separate connection to the Bitcoin RPC server.
```
ceyhunsen, line 427: This isn't incorrect but we need this because inner RPC struct is not clonable nor copiable. So, we need to create a new client to clone `ExtendedRpc`.

ceyhunsen (changes requested):
Looks like we are going to have an RPC endpoint for debugging TXs. Isn't this going to increase disk usage significantly? If so, some verifiers might not want to have this functionality. Maybe we can make this code work only if a specific feature flag is given instead of my recommendation (`#[cfg(test)]`).

Haven't checked tx sender module, yet.

core/src/extended_rpc.rs
mmtftr: cloning = opening new client connection here, so I don't see what's wrong with the docs.

core/src/database/tx_sender.rs
mmtftr: this is tx debugging info for production environments. we need this so we can show the operator what went wrong when sending a tx. It'll be useful alongside the usage of log filtering based on try_to_send_id

core/src/extended_rpc.rs
ceyhunsen: Nothing wrong. Just wanted to point out the real reason why we have this. I was also only replying to second sentence.

core/src/database/tx_sender.rs
```diff
@@ -107,6 +109,19 @@ impl Database {
         .execute(tx.deref_mut())
         .await?;
 
+        // Get confirmed direct transactions for debugging
+        let confirmed_direct_txs: Vec<(i32, TxidDB)> = sqlx::query_as(&format!(
```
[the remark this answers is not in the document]
mmtftr: I do agree that this adds a lot of overhead to this function and should be refactored to a bg task if possible

core/src/database/tx_sender.rs
```diff
@@ -541,8 +621,164 @@ impl Database {
                 .map(u32::try_from)
                 .transpose()
                 .wrap_err("Failed to convert seen_block_id to u32")?,
+            serde_json::from_str(result.4.as_deref().unwrap_or("null")).wrap_err_with(|| {
+                format!("Failed to decode rbf_signing_info from {:?}", result.4)
+            })?,
         ))
```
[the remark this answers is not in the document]
mmtftr: I think this already behaves as expected by handling missing info as Option::none, I'll recheck with a test.

mmtftr:
I don't think it'll increase disk usage by much. I'm not storing historical data of all txs but just the latest state of each tx and submission errors for any tx. I think submission errors can quickly grow if there is a big problem, so we might want to limit the number of submission errors per tx.

core/src/database/tx_sender.rs
mmtftr: I meant Tx Debugging functions not test-only or "compiled in debug" functions though.

core/src/database/tx_sender.rs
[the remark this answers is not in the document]
mmtftr: OK, I refactored this part into a bg task to ensure minimal effect on core behavior.

core/src/tx_sender/rbf.rs
```diff
@@ -0,0 +1,1357 @@
...
+        &self,
+        txid: &Txid,
+        new_feerate: FeeRate,
+    ) -> Result<Option<Amount>> {
+        let original_tx = self.rpc.get_tx_of_txid(txid).await.map_err(|e| eyre!(e))?;
+
+        // Calculate original tx fee
+        let original_tx_fee = self.get_tx_fee(&original_tx).await.map_err(|e| eyre!(e))?;
+
+        println!("original_tx_fee: {}", original_tx_fee);
```
ceyhunsen, line 50:
```suggestion
        tracing::info!("original_tx_fee: {}", original_tx_fee);
```

ekrembal (approved):

Issue #538, opened 2025-02-14
Get N-of-N aggregated public key endpoint for aggregator


ekrembal:
This will make our tests much easier, we will be able to generate the correct deposit address

ceyhunsen:
#661

PR #661, opened 2025-04-04
Additional cli functionality and bug fixes

ekrembal:


core/src/rpc/clementine.proto
```diff
@@ -13,6 +13,15 @@ message Outpoint {
   uint32 vout = 2;
 }
 
+message XonlyPublicKey {
```
ceyhunsen, line 16: There is also `VerifierPublicKeys`

core/src/rpc/clementine.proto
```diff
@@ -483,4 +494,6 @@ service ClementineAggregator {
 
   // send a pre-signed tx
   rpc InternalSendTx(SendTxRequest) returns (Empty) {}
+
+  rpc GetNofNAggregatedXonlyPk(Empty) returns (NofNResponse) {}
```
ceyhunsen, line 498: Do you think we need this endpoint? We can fetch this info also via `setup()`. Maybe we should choose which approach we should go with for these kinds of testing/setup cases and not duplicate solutions.

I think we can adopt this approach in tests also, instead of relying on `setup()`. If this is easier for deployment tests, let's remove `setup()`'s return value.

ceyhunsen (changes requested):
It looks OK for itself but duplicates existing stuff.

Also, didn't test changes myself. But I think you already did.

ceyhunsen:
Related #538

core/src/verifier.rs
```diff
@@ -79,6 +79,18 @@ impl NofN {
...
+
+    pub fn new_without_idx(
+        public_keys: Vec<bitcoin::secp256k1::PublicKey>,
+    ) -> Result<Self, BridgeError> {
+        let agg_xonly_pk =
+            bitcoin::secp256k1::XOnlyPublicKey::from_musig2_pks(public_keys.clone(), None)?;
+        Ok(NofN {
+            public_keys,
+            agg_xonly_pk,
+            idx: 0,
```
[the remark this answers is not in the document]
ceyhunsen: With #662, this will be removed. So, don't bother.

core/src/bin/cli.rs
```diff
@@ -241,22 +266,150 @@ async fn handle_aggregator_call(url: String, command: AggregatorCommands) {
...
+                    .get_ref()
+                    .nofn_xonly_pk
+                    .clone(),
+            };
+
+            let num_verifiers = match num_verifiers {
+                Some(num) => num,
+                None => {
+                    aggregator
+                        .get_nof_n_aggregated_xonly_pk(Request::new(Empty {}))
```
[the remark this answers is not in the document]
ekrembal: it's cli so it's ok

core/src/rpc/clementine.proto
ekrembal: That is different?

core/src/rpc/clementine.proto
ekrembal: Yes, let's have it but then we can remove after Atacan's pr

ceyhunsen (approved):

core/src/bin/cli.rs
```diff
@@ -241,22 +266,150 @@ async fn handle_aggregator_call(url: String, command: AggregatorCommands) {
...
+        }
+        AggregatorCommands::GetDepositAddress {
+            evm_address,
+            recovery_taproot_address,
+            network,
+            bridge_amount,
+            user_takes_after,
+        } => {
+            let response = aggregator
+                .get_nof_n_aggregated_xonly_pk(Request::new(Empty {}))
```
atacann, line 356: We probably still need get_nofn even after my PR because of GetDepositAddress, but it should be fine. 

Also maybe change call name to get_nofn_aggregated_xonly_pk.

ekrembal: that came from proto i guess, fixing it rn

Issue #478, opened 2025-01-28
Enhance the indexing of the entities


ozankaymak:
### Proposal Description

Right now, there seems to be a problem inside the integration tests. We did not realize it until now but when implementing **aggregator_setup_watchtower_challenge_addresses**, I came across this problem. As the tests get more and more complicated, we need to make sure everything works as intended.

ceyhunsen:
#662

Issue #544, opened 2025-02-17
Get rid of operator_idx, use xonlypk everywhere


ekrembal:

ceyhunsen:
#662

PR #662, opened 2025-04-07
Atacan/remove indexes

atacann:
# Description

## Linked Issues

- Closes # (issue, if applicable)
- Related to # (issue)

## Testing

added unit tests?

## Docs

have the docs been updated?

### DIFF
--- a/core/src/aggregator.rs
+++ b/core/src/aggregator.rs
@@ -18,6 +18,7 @@ use crate::{
 };
 use bitcoin::hashes::Hash;
 use bitcoin::secp256k1::{schnorr, Message, PublicKey};
+use bitcoin::XOnlyPublicKey;
 use futures_util::future::try_join_all;
 use secp256k1::musig::{MusigAggNonce, MusigPartialSignature};
 use std::sync::Arc;
@@ -37,9 +38,10 @@ pub struct Aggregator {
     pub(crate) db: Database,
     pub(crate) config: BridgeConfig,
     pub(crate) tx_sender: TxSenderClient,
-    pub(crate) operator_clients: Vec<ClementineOperatorClient<tonic::transport::Channel>>,
+    operator_clients: Vec<ClementineOperatorClient<tonic::transport::Channel>>,
     verifier_clients: Vec<ClementineVerifierClient<tonic::transport::Channel>>,
     verifier_keys: Arc<tokio::sync::RwLock<Vec<PublicKey>>>,
+    operator_keys: Arc<tokio::sync::RwLock<Vec<XOnlyPublicKey>>>,
 }
 
 impl Aggregator {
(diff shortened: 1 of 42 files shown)


core/src/aggregator.rs
```diff
@@ -269,4 +288,46 @@ impl Aggregator {
 
         Ok(final_sig)
     }
+
+    /// Returns a list of verifier clients that are participating in the deposit.
+    pub async fn get_participating_verifiers(
```
ceyhunsen, line 293: maybe we can have a third private function that can accept the 2 required inputs and make these public functions as wrappers. this way we can reduce code duplication.

core/src/operator.rs
```diff
@@ -592,7 +585,7 @@ where
                     txhandler.get_transaction_type()
                 {
                     let partial = PartialSignatureInfo {
-                        operator_idx: self.idx,
+                        operator_idx: 0, // dummy value
```
ceyhunsen, line 588: does this doesnt matter or left as a todo?

core/src/rpc/clementine.proto
```diff
@@ -84,35 +84,40 @@ message WinternitzPubkey {
...
-    BaseDeposit base_deposit = 1;
-    ReplacementDeposit replacement_deposit = 2;
+    BaseDeposit base_deposit = 2;
+    ReplacementDeposit replacement_deposit = 3;
   }
 }
 
+message Actors {
+  /// Public keys of verifiers that will participate in the deposit.
+  repeated bytes verifiers = 1;
```
ceyhunsen, line 102: there should be another type called `VerifierPublicKeys`. It can be changed to just public keys and a new type for x-only pk can be created to be used here. It will be easier to write `try_from` impls for this after that.

core/src/rpc/clementine.proto
```diff
@@ -259,6 +263,10 @@ message FinalizedPayoutParams {
   Outpoint deposit_outpoint = 2;
 }
 
+message XOnlyPublicKeyRpc {
```
ceyhunsen, line 266: so there is a message for xonly pk but not used everywhere. is there a reason for this?

ceyhunsen (commented):
naturally lots of lines has changed and i might miss some stuff. but to me it looks ok mostly.

ceyhunsen:
Related #544

ceyhunsen:
Could close #478

core/src/operator.rs
atacann: Doesn't matter, will clarify comment

core/src/rpc/clementine.proto
atacann: 👍 Changed them in [5a8ad18](https://github.com/chainwayxyz/clementine/pull/662/commits/5a8ad184a7f50fe7d56fffcc49452f1ff4c285f7)

core/src/rpc/clementine.proto
ceyhunsen: Why not `XOnlyPublicKeys` instead?

ceyhunsen (approved):
Small suggestions left but they can be omitted for now

Issue #658, opened 2025-04-03
Challenge extra checks


atacann:
For normal challenge
- Check if operators burn connector for the round is still on-chain, because otherwise a challenge is not needed.
- Check if operator was already challenged in the same round, to not send multiple challenges to the same round.

For watchtower challenge
- Check if a challenge for the same deposit-operator pair was already sent, if so don't create a new challenge or send the same challenge.

PR #666, opened 2025-04-09
Challenge checks

atacann:
#658 

- Check if operators burn connector for the round is still on-chain, because otherwise a challenge is not needed.

Now round state machine checks if operator collateral was spent in an unexpected way (meaning not next round or next ready to reimburse tx), if so it goes to operator_exit state and does nothing anymore (doesn't check for new kickoffs). As new kickoffs are checked with state machine, if round machine state is operator_exit even if operator sends a malicious kickoff it will be ignored.

Already existing kickoff state machine already checked if burn connector was spent and close the state if so, so no need to change that.

- Check if operator was already challenged in the same round, to not send multiple challenges to the same round.

Round machine state now stores if after checking a kickoff it was challenged, and gives this information to consequent new kickoff checks. So after verifier finds a kickoff is malicious (these kickoffs are checked in their tx order in the block), subsequent checkkickoffs will know there was a challenge before and not send challenge tx.


core/src/builder/transaction/creator.rs
```diff
@@ -144,8 +146,12 @@ impl ReimburseDbCache {
                 self.operator_data = Some(
                     self.db
                         .get_operator(None, self.operator_idx as i32)
-                        .await?
-                        .ok_or(BridgeError::OperatorNotFound(self.operator_idx))?,
+                        .await
+                        .map_err(|e| eyre!("Failed to get operator data from database: {}", e))?
```
mmtftr, line 150: wrong usage, please do a wrap_err with the static string instead

```suggestion
                        .wrap_err("Failed to get operator data from database")?
```

core/src/builder/transaction/creator.rs
```diff
@@ -176,7 +184,10 @@ impl ReimburseDbCache {
                 self.kickoff_winternitz_keys = Some(KickoffWinternitzKeys::new(
                     self.db
                         .get_operator_kickoff_winternitz_public_keys(None, self.operator_idx)
-                        .await?,
+                        .await
+                        .map_err(|e| {
```
mmtftr, line 188: wrap_err to preserve error chain as above

core/src/builder/transaction/creator.rs
```diff
@@ -195,10 +206,11 @@ impl ReimburseDbCache {
                     let (assert_addr, bitvm_hash) = self
                         .db
                         .get_bitvm_setup(None, self.operator_idx as i32, *deposit_outpoint)
-                        .await?
-                        .ok_or(TxError::BitvmSetupNotFound(
-                            self.operator_idx as i32,
-                            deposit_outpoint.txid,
+                        .await
+                        .map_err(|e| eyre!("Failed to get BitVM setup from database: {}", e))?
```
mmtftr, line 210: wrap_err to preserve error chain as above

core/src/builder/transaction/creator.rs
```diff
@@ -245,10 +259,11 @@ impl ReimburseDbCache {
                     let bitvm_hash = self
                         .db
                         .get_bitvm_root_hash(None, self.operator_idx as i32, *deposit_outpoint)
-                        .await?
-                        .ok_or(TxError::BitvmSetupNotFound(
-                            self.operator_idx as i32,
-                            deposit_outpoint.txid,
+                        .await
+                        .map_err(|e| eyre!("Failed to get BitVM root hash from database: {}", e))?
```
mmtftr, line 263: wrap_err to preserve error chain as above

core/src/states/mod.rs
```diff
@@ -438,12 +438,16 @@ impl<T: Owner + std::fmt::Debug + 'static> StateManager<T> {
...
     /// # Errors
-    /// If the state machines do not stabilize after 50 iterations, we return an error.
+    /// If the state machines do not stabilize after some iterations, we return an error.
     pub async fn process_block_parallel(&mut self, block_height: u32) -> Result<(), eyre::Report> {
         eyre::ensure!(
             self.context.cache.block_height == block_height,
             "Block cache is not updated"
         );
+
+        let saved_kickoff_machines = self.kickoff_machines.clone();
```
mmtftr, line 448: machines_checkpoint

core/src/states/mod.rs
```diff
@@ -505,9 +517,9 @@ impl<T: Owner + std::fmt::Debug + 'static> StateManager<T> {
                 changed_kickoff_machines.extend(std::mem::take(&mut ctx.new_kickoff_machines));
             }
 
-            if iterations > 500 {
+            if iterations > 3000 {
```
mmtftr, line 520: set to 10000 and add // TODO: https://github.com/chainwayxyz/clementine/issues/675

core/src/states/context.rs
```diff
@@ -86,6 +87,12 @@ pub enum Duty {
     },
 }
 
+/// Result of handling a duty
+#[derive(Debug, Clone, Default)]
+pub struct DutyResult {
```
mmtftr, line 92: I feel like this is better as an enum. Though, I'm not sure if we're gonna add anything else to it later on or if we'll need to combine multiple enum items.

core/src/test/full_flow.rs
```diff
@@ -799,11 +799,13 @@ pub async fn run_challenge_with_state_machine(
     rpc.mine_blocks(config.protocol_paramset().time_to_send_watchtower_challenge as u64)
         .await?;
 
+    tracing::warn!("Checking if watchtower challenge utxos were spent");
```
mmtftr, line 802: remove unnecessary warn logs (likely remaining from debugging)

core/src/operator.rs
```diff
@@ -342,6 +343,15 @@ where
     /// Creates the round state machine by adding a system event to the database
     pub async fn track_rounds(&self) -> Result<(), BridgeError> {
         let mut dbtx = self.db.begin_transaction().await?;
+        // set operators own kickoff winternitz public keys before creating the round state machine
+        // as round machine needs kickoff keys to create the first round tx
```
mmtftr, line 347: question: where do we actually send the first round tx? couldn't find it.

mmtftr (commented):
Thanks, some comments about behavior and code style

core/src/operator.rs
atacann: It is in Operator::get_params 😄 . I added it to https://github.com/chainwayxyz/clementine/issues/667 now. The kickoff wots should also be saved in db after its own operators get_params reaches verifier (without this change).

ekrembal (approved):

ceyhunsen (approved):

mmtftr (approved):

Issue #660, opened 2025-04-03
Check nofn correctness


atacann:
When a new deposit request comes, check if nofn in citrea contract matches

atacann:
#669

PR #669, opened 2025-04-10
Check nofn key

atacann:
#660 

Adds a function to CitreaClientT to check if a xonly pk matches nofn_xonly_pk written in the citrea contract. 
Mock client always returns true, CitreaClient retrieves nofn key from contracts scriptPrefix variable and checks equality.

Also re-enables citrea e2e test that was ignored.

Note: As contract is getting updated we need to update check_nofn_correctness function and Bridge.json file later.

### DIFF
--- a/core/src/verifier.rs
+++ b/core/src/verifier.rs
@@ -434,6 +434,8 @@ where
         session_id: u32,
         mut agg_nonce_rx: mpsc::Receiver<MusigAggNonce>,
     ) -> Result<mpsc::Receiver<MusigPartialSignature>, BridgeError> {
+        self.verify_deposit_nofn(&deposit_data).await?;
+
         let verifier = self.clone();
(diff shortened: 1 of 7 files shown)


core/src/verifier.rs
```diff
@@ -1224,6 +1228,17 @@ where
         dbtx.commit().await?;
         Ok(())
     }
+
+    async fn verify_deposit_nofn(&self, deposit_data: &DepositData) -> Result<(), BridgeError> {
```
ceyhunsen, line 1232: Same function in both operator and verifier. Maybe we can make `check_nofn_correctness` return `Ok(())` and remove these functions?

ceyhunsen (changes requested):

ceyhunsen (approved):

ekrembal (approved):

PR #680, opened 2025-04-15
Deposit with same input

ceyhunsen:
# Description

Updates on conflict rules for set_deposit_signatures.

### DIFF
--- a/core/src/database/operator.rs
+++ b/core/src/database/operator.rs
@@ -386,7 +386,9 @@ impl Database {
         let query = sqlx::query(
             "
             INSERT INTO deposit_signatures (deposit_id, operator_xonly_pk, round_idx, kickoff_idx, kickoff_txid, signatures)
-            VALUES ($1, $2, $3, $4, $5, $6);"
+            VALUES ($1, $2, $3, $4, $5, $6)
+            ON CONFLICT (deposit_id, operator_xonly_pk, round_idx, kickoff_idx)
+            DO UPDATE SET kickoff_txid = EXCLUDED.kickoff_txid, signatures = EXCLUDED.signatures;"
         )
         .bind(i32::try_from(deposit_id).wrap_err("Failed to convert deposit id to i32")?)
         .bind(XOnlyPublicKeyDB(operator_xonly_pk))
@@ -396,6 +398,7 @@ impl Database {
         .bind(SignaturesDB(DepositSignatures{signatures}));
 
         execute_query_with_tx!(self.connection, tx, query, execute)?;
+
         Ok(())
     }
 
--- a/core/src/database/operator.rs
+++ b/core/src/database/operator.rs
@@ -20,7 +20,7 @@ use crate::{
 use bitcoin::{OutPoint, Txid, XOnlyPublicKey};
 use bitvm::signatures::winternitz;
 use bitvm::signatures::winternitz::PublicKey as WinternitzPublicKey;
-use eyre::Context;
+use eyre::{eyre, Context};
 use std::str::FromStr;
 
 pub type RootHash = [u8; 32];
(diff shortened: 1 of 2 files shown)


aoengin (approved):
lgtm

core/src/database/operator.rs
```diff
@@ -386,7 +386,9 @@ impl Database {
         let query = sqlx::query(
             "
             INSERT INTO deposit_signatures (deposit_id, operator_xonly_pk, round_idx, kickoff_idx, kickoff_txid, signatures)
-            VALUES ($1, $2, $3, $4, $5, $6);"
+            VALUES ($1, $2, $3, $4, $5, $6)
+            ON CONFLICT (deposit_id, operator_xonly_pk, round_idx, kickoff_idx)
+            DO UPDATE SET kickoff_txid = EXCLUDED.kickoff_txid, signatures = EXCLUDED.signatures;"
```
mmtftr, line 391: was this causing a problem? also what does it mean if we try to override these signatures? will we lose the correct signatures when another deposit is triggered with the same ID?

core/src/rpc/aggregator.rs
```diff
@@ -1198,4 +1198,101 @@ mod tests {
...
+            .try_into()
+            .unwrap();
+        let movetx_txid: Txid = aggregator
+            .new_deposit(clementine::Deposit::from(deposit_info))
+            .await
+            .unwrap()
+            .into_inner()
+            .try_into()
+            .unwrap();
+        rpc.mine_blocks(1).await.unwrap();
```
mmtftr, line 1269: I think we should decide on the behavior of the bridge when a double deposit is attempted, then also assert this behavior in the test.

core/src/database/operator.rs
ceyhunsen: This was causing an error about uniqueness when second deposit call is issued with same params. With this change, if a combination of (deposit_id, operator_xonly_pk, round_idx, kickoff_idx) is present in db, their (kickoff_txid, signatures) will be replaced with the latest call. Not sure if this is the correct behavior though; We can do nothing on conflict, too.

atacann: In this case it doesn't matter if we update or do nothing as both new and old signatures should be correct. I think in general also for get_params etc we should do nothing if it conflicts, so that for example if an operator somehow changes their kickoff wots keys we don't overwrite it in our db, because if we overwrite we cannot generate old signed kickoff txs. If an operator/verifier wants to change something about their xonly key/wots key, they need to create a new operator or verifier.

ozankaymak: I think in terms of security measures we better re-sign the whole thing, since one can tweak some small details and if the verifiers use the same nonces again, they will compromise their private keys.

ceyhunsen: Than neither of the options should be viable, right? If altered `kickoff_txid` and `signatures` are given, we shouldn't silently ignore this conflict and throw an error, instead of doing nothing. This will require additional code, instead of just changing the query.

atacann: For each deposit operators share new wots keys (for our current code they will share same wots keys if deposit outpoint is the same but it can be changed by editing code). With different wots keys txids will change so if we use same nonce (idk if we do) and on conflict update wots keys in db, their keys will get leaked.

ozankaymak: Can't we just add some randomness to verifiers' nonce generation?

ceyhunsen: https://github.com/chainwayxyz/clementine/pull/680/commits/52b9206c6413c6e97f603ea6db1aa232e6804fb3

ekrembal (approved):

PR #685, opened 2025-04-16
Fix tx sender fee rate getting & logs

ekrembal:


mmtftr:
Why do we need these info logs? Can they not be debug? These are very fine-grained logs that are only going to be used by someone debugging the txsender module.

ekrembal:
@mmtftr let's merge this like this for now. we can clear logs later

PR #692, opened 2025-04-18
Replace Deposit & Bug fixes

ekrembal:


core/src/bitcoin_syncer.rs
```diff
@@ -75,6 +75,12 @@ pub(crate) async fn save_block(
         block_height
     );
 
+    // update the block_info as canonical if it already exists
+    let block_id = db.set_block_as_canonical_if_exists(Some(dbtx), block_hash).await?;
```
mmtftr, line 79: why is this required btw? is this the case of a reorg happening to a previously discarded alternate chain?

mmtftr: also, if this is the case, shouldn't we re-process the transactions in the db?

core/src/database/tx_sender.rs
```diff
@@ -141,41 +141,65 @@ impl Database {
         tx: DatabaseTransaction<'_, '_>,
         block_id: u32,
     ) -> Result<(), BridgeError> {
-        // Unconfirm tx_sender_fee_payer_utxos
```
mmtftr, line 144: what's the difference of executing these separately again?

core/src/states/task.rs
```diff
@@ -113,8 +113,21 @@ impl<T: Owner + std::fmt::Debug + 'static> Task for BlockFetcherTask<T> {
...
+                        tracing::error!("Block at height {} not found in BlockFetcherTask, current tip height is {}", self.next_height, current_tip_height);
+                        return Err(BridgeError::Error(format!(
+                            "Block at height {} not found in BlockFetcherTask, current tip height is {}",
+                            self.next_height, current_tip_height
+                        )));
+                    }
+
                     let event = SystemEvent::NewBlock {
-                        block_id,
+                        block_id: new_block_id.expect("it must be some"),
```
mmtftr, line 130: rather do

```rs
let Some(block_id) = new_block_id else {
    tracing::error!("Block at height {} not found in BlockFetcherTask, current tip height is {}", self.next_height, current_tip_height);
                        return Err(BridgeError::Error(format!(
                            "Block at height {} not found in BlockFetcherTask, current tip height is {}",
                            self.next_height, current_tip_height
                        )));
}
```
above

core/src/database/verifier.rs
```diff
@@ -12,6 +12,38 @@ use eyre::Context;
 use sqlx::QueryBuilder;
 
 impl Database {
+    /// Returns the last deposit index.
+    /// If no deposits exist, returns None
+    pub async fn get_last_deposit_idx(
+        &self,
+        tx: Option<DatabaseTransaction<'_, '_>>,
+    ) -> Result<Option<u32>, BridgeError> {
+        let query = sqlx::query_as::<_, (i32,)>("SELECT COALESCE(MAX(idx), -1) FROM withdrawals");
```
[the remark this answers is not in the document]
ceyhunsen: Maybe we should rename this as `citrea_deposits_and_withdrawals`

ceyhunsen (dismissed):

ceyhunsen (approved):

PR #706, opened 2025-04-23
Proving

atacann:
test to get actual proofs for circuit tests


core/src/test/common/tx_utils.rs
```diff
@@ -146,7 +146,7 @@ pub async fn ensure_tx_onchain(rpc: &ExtendedRpc, tx: Txid) -> Result<(), eyre::
             let _ = mine_once_after_in_mempool(rpc, tx, Some("ensure_tx_onchain"), Some(1)).await;
             Ok(false)
         },
-        None,
+        Some(Duration::from_secs(5000)),
```
ozankaymak, line 149: What is this for?

core/src/header_chain_prover.rs
```diff
@@ -0,0 +1,853 @@
...
+        let env = env
+            .build()
+            .map_err(|e| eyre::eyre!(e))
+            .wrap_err("Failed to build environment")?;
+
+        let prover = risc0_zkvm::default_prover();
+
+        let elf = match self.network {
+            Network::Bitcoin => MAINNET_WORK_ONLY_ELF,
+            Network::Testnet => TESTNET4_WORK_ONLY_ELF,
```
ozankaymak, line 322: We are not using Testnet - this is Testnet3. We do not need this.

core/src/test/common/tx_utils.rs
atacann: Tests end too early otherwise because proving takes time

PR #715, opened 2025-04-25
TxSender Documentation and Reorg Tests

### LINKED ISSUE #699: Reorg tests

ceyhunsen:
# Description

- Adds documentation to some parts of the `tx_sender`
- Adds reorg tests
  - These test are failing because CPFP packages loses it's child tx
  - CPFP packages needed to be re-created after a reorg
  - Fee payer TXs are OK
  - Tests are ignored till' this is fixed

## Linked Issues

- Closes #699


ceyhunsen:
From https://gnusha.org/pi/bitcoindev/CAFXO6=JROe_9ih2h+_CCH-UbxehsM5RQ6YyNnPesEpveBEtdow@mail.gmail.com/:

> A child-with-unconfirmed-parents package is defined for a transaction
> based on the current chain state. As such, a new block extending the
> tip may decrease the number of transactions in the package (i.e. if
> any of the transaction's parents were included in the block). In a
> reorg, the number of transactions in the package may decrease or
> increase (i.e. if any of the transaction's parents were included in a
> block in the previous chain but not the new one).

As far as I understood If a reorg unconfirms both the fee_payer tx and the cpfp package, cpfp package's child gets discarded because fee_payer tx is not confirmed.

core/src/database/tx_sender.rs
```diff
@@ -320,17 +317,34 @@ impl Database {
...
     }
 
+    /// Returns all unconfirmed fee payer transactions for a given bumped
+    /// transaction.
+    ///
     /// Some fee payer TXs may not hit onchain, so we need to bump their fees.
-    /// These TXs should not be confirmed and should not be replaced by other TXs.
-    /// Replaced means that the TX was bumped and the replacement TX is in the database.
-    pub async fn get_bumpable_fee_payer_txs(
+    /// These TXs should not be confirmed and should not be replaced by other
```
mmtftr, line 331: This explanation is ambiguous and includes unnecessary information, let's delete this paragraph and incorporate it to the top. The following should be clear enough

```
Returns all unconfirmed fee payer transactions for a try-to-send Tx. Replaced (bumped) fee payers are not included. 
```

core/src/tx_sender/mod.rs
```diff
@@ -1,11 +1,40 @@
...
+//! fees and making sure that transactions are finalized until the deadline, all
+//! in background.
+//!
+//! ## Concepts
+//!
+//! ### Transaction/UTXO Activation and Cancelation
+//!
+//! Transactions and UTXOs can be marked as active, non-active or cancelled.
+//!
+//! Transactions/UTXOS that are marked as active are candidate transactions for
```
mmtftr, line 13: what does it mean for a UTXO to be active?

mmtftr: activation is only used for transactions, not UTXOs. UTXOs can be spent which trigger a condition.

core/src/tx_sender/mod.rs
```diff
@@ -1,11 +1,40 @@
...
+//! transaction sender. Non-active transactions/UTXOS on the other hand are, not
+//! yet processed by the transaction sender.
+//!
+//! If a transaction/UTXO is marked as cancelled, it can't be used in
+//! transaction sender anymore. Spent transactions/UTXOs are examples for that.
+//!
+//! An active transaction can be send if its UTXOs are also marked as active
```
mmtftr, line 20: I don't like this module doc because it uses very ambiguous wording. What does it mean if a "Transaction/UTXO can't be used in transaction sender any more"? What is a "candidate transaction"? What do you mean by "processed by the transaction sender"? By UTXO, which UTXO are we talking about, fee payer UTXOs or cancellation outpoints or activation outpoints?

The TxSender can "use"/"process" transactions in a couple of ways. It can "try to send" them which involves (for CPFP) "creating fee payer UTXOs" to satisfy the target fee rate and "sending the Tx" by way of CPFP (ie. `submit_package`) or (for RBF) "creating fee payer UTXOs" and "sending the Tx" by way of RBF (ie. add funding, resign the transaction, sendrawtransaction). It can also "bump a Tx" by way of `bump_fee`ing the RBF or re-sending the CPFP package with larger fees.

I think it'd be much better if we can define and develop this domain-specific terminology in the module documentation, before we talk about the behavior of the module and dive into specifics.

mmtftr (changes requested):

core/src/database/tx_sender.rs
```diff
@@ -785,7 +815,7 @@ impl Database {
     ) -> Result<Vec<(Txid, u32, Amount, bool)>, BridgeError> {
         let query = sqlx::query_as::<_, (TxidDB, i32, i64, Option<i32>)>(
             r#"
-            SELECT fee_payer_txid, vout, amount, seen_block_id IS NOT NULL as confirmed
+            SELECT fee_payer_txid, vout, amount, (seen_block_id IS NOT NULL)::INT4 as confirmed
```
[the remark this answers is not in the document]
ceyhunsen: `fetch_optional` returns only the first row. So, that will change the behavior. I added this because when `seen_block_id` is null, it will default to the type `bool` and throw error about types. Do you mean anything else?

core/src/database/tx_sender.rs
ceyhunsen: https://github.com/chainwayxyz/clementine/pull/715/commits/b736a52b77239049cddc2eefbc25ce22717801bf

core/src/database/tx_sender.rs
```diff
@@ -516,6 +530,22 @@ impl Database {
         Ok(())
     }
 
+    /// Returns sendable transactions that are not in the non-active list and
```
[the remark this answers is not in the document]
ceyhunsen: https://github.com/chainwayxyz/clementine/pull/715/commits/b736a52b77239049cddc2eefbc25ce22717801bf because there is still the fee_rate argument, i hesitated to delete that comment and added deprecated warning instead

core/src/database/tx_sender.rs
ceyhunsen: forgot to add it to the top. will add

ceyhunsen:
this is going to get replaced by the #852 and one other pr. reorg tests in this pr are valuable and will be reused.

PR #721, opened 2025-04-29
Integrate header-chain and some elements of final-spv

ozankaymak:


Cargo.toml
```diff
@@ -5,6 +5,7 @@ members = [
   "circuits-lib",
   "risc0-circuits/bridge-circuit",
   "risc0-circuits/work-only",
+  "risc0-circuits/header-chain",
```
aoengin, line 8: lets keep guests isolated

bridge-circuit-host/src/mock_zkvm.rs
```diff
@@ -0,0 +1,51 @@
...
+        let data = self.data.lock().unwrap();
+        T::try_from_slice(&data.values).unwrap()
+    }
+    fn commit<T: borsh::BorshSerialize>(&self, item: &T) {
+        let mut data = self.data.lock().unwrap();
+        let value = borsh::to_vec(item).unwrap();
+        data.journal.extend_from_slice(&value);
+    }
+
+    fn verify<T: borsh::BorshSerialize>(&self, _method_id: [u32; 8], _journal: &T) {}
```
aoengin, line 33: Should we add a warning message to let people know this method will not actually perform a verification?

bridge-circuit-host/src/mock_zkvm.rs
```diff
@@ -0,0 +1,51 @@
...
+    }
+    fn prove(&self, _elf: &[u32]) -> Proof {
+        let data = self.data.lock().unwrap();
+        Proof {
+            method_id: [42; 8],
+            journal: data.journal.clone(),
+        }
+    }
+
+    fn add_assumption(&self, _proof: Proof) {}
```
aoengin, line 50: same as above, we can add a warning message.

circuits-lib/src/bridge_circuit/merkle_tree.rs
```diff
@@ -0,0 +1,195 @@
+use borsh::{BorshDeserialize, BorshSerialize};
```
aoengin, line 1: I think we can create a separate folder for all files that are not directly related to the guest. Maybe like `helper`?

circuits-lib/src/header_chain/mmr_guest.rs
```diff
@@ -0,0 +1,94 @@
...
+        while size % 2 == 1 {
+            let sibling = self.subroots.pop().unwrap();
+            current = hash_pair(sibling, current);
+            size /= 2
+        }
+        self.subroots.push(current);
+        self.size += 1;
+    }
+
+    // fn get_helpers_from_index(&self, index: u32) -> (usize, usize, u32) {
```
aoengin, line 43: We can remove the commented code to keep it clean If we do not plan to use it in the future.

risc0-circuits/work-only/build.rs
```diff
@@ -28,8 +28,8 @@ fn main() {
     }
 
     let network = env::var("BITCOIN_NETWORK").unwrap_or_else(|_| {
-        println!("cargo:warning=BITCOIN_NETWORK not set, defaulting to 'mainnet'");
-        "mainnet".to_string()
+        println!("cargo:warning=BITCOIN_NETWORK not set, defaulting to 'testnet4'");
+        "testnet4".to_string()
```
aoengin, line 32: default should be mainnet

risc0-circuits/work-only/build.rs
```diff
@@ -149,8 +149,8 @@ fn copy_binary_to_elfs_folder(network: String) {
         "signet" => "../elfs/signet-work-only-guest.bin",
         "regtest" => "../elfs/regtest-work-only-guest.bin",
         _ => {
-            println!("cargo:warning=Invalid network specified, defaulting to mainnet");
-            "../elfs/mainnet-work-only-guest.bin"
+            println!("cargo:warning=Invalid network specified, defaulting to testnet4");
+            "../elfs/testnet4-work-only-guest.bin"
```
aoengin, line 153: same

aoengin (changes requested):

bridge-circuit-host/src/mock_zkvm.rs
ozankaymak: This is only for testing purposes, so I guess we do not need to

risc0-circuits/work-only/build.rs
ozankaymak: This is again for testing purposes, otherwise we need to pass an environment variable like `BITCOIN_NETWORK=... cargo test ...`

circuits-lib/src/header_chain/mmr_guest.rs
ozankaymak: We will get rid of MMRs in the future so that is a concern for a seperate PR

aoengin (approved):

PR #727, opened 2025-05-01
Update Create SPV

aoengin:


atacann (approved):
Lgtm, blockhash can also be BlockHash type of bitcoin lib but no big deal.

PR #737, opened 2025-05-03
Emergency stop Tx

ekrembal:
Collecting Emergency stop Tx signature to move funds to a security council


mmtftr:
These PRs are sequential, will review this after #730 is merged

core/src/rpc/aggregator.rs
```diff
@@ -583,6 +616,61 @@ impl Aggregator {
...
+            &verifiers_public_keys,
+            None,
+            emergency_stop_agg_nonce,
+            &musig_partial_sigs,
+            Message::from_digest(sighash.to_byte_array()),
+        )
+        .map_err(|x| {
+            BridgeError::Error(format!(
+                "Aggregating Emergency Stop signatures failed {}",
+                x
```
mmtftr, line 657: this discards the error chain information, prefer `wrap_err("Failed to aggregate emergency stop signatures")` (you may need to import eyre::Context)

core/src/tx_sender/mod.rs
```diff
@@ -314,7 +314,7 @@ impl TxSender {
         new_fee_rate: FeeRate,
         current_tip_height: u32,
     ) -> Result<()> {
-        tracing::info!("Trying to send unconfirmed txs with new fee rate: {new_fee_rate:?}, current tip height: {current_tip_height:?}");
+        // tracing::info!("Trying to send unconfirmed txs with new fee rate: {new_fee_rate:?}, current tip height: {current_tip_height:?}");
```
mmtftr, line 317: convert to tracing::debug instead?

core/src/verifier.rs
```diff
@@ -684,16 +703,60 @@ where
...
 
+        let emergency_stop_secnonce = {
+            let mut session_map = self.nonces.lock().await;
+            let session = session_map
+                .sessions
+                .get_mut(&session_id)
+                .ok_or_else(|| eyre::eyre!("Could not find session id {session_id}"))?;
+            session
+                .nonces
+                .pop()
```
mmtftr, line 714: Isn't the order of movetx and emergency_stop reversed here?

This pops movetx secnonce first, then emergency secnonce second. So in the array it's supposed to be

`secnonces = [...other nonces, emergency, movetx]`.

I would've expected the opposite since the nonce aggregator aggregates movetx first then emergency stop tx. So I don't understand how the tests don't fail here.

This ordering is quite arbitrary, maybe we should consider using SignatureId to store a HashMap<SignatureId, Nonce> which would make this less prone to failure. (ofc out of scope for this PR)

mmtftr (changes requested):

core/src/verifier.rs
ekrembal: if the order is incorrect it gives error

core/src/tx_sender/mod.rs
ceyhunsen: Actually, it is better to delete this log because it clutters logs too much. We have similar logs already in the following lines.

core/src/rpc/aggregator.rs
ekrembal: 6f3eb32

core/src/verifier.rs
```diff
@@ -439,12 +440,12 @@ where
                 }
             }
 
-            let last_nonce = session
-                .nonces
-                .pop()
-                .ok_or(eyre::eyre!("No last nonce available"))?;
-            session.nonces.clear();
-            session.nonces.push(last_nonce);
+            assert_eq!(
```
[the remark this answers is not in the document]
mmtftr: need to remove panic before merge

core/src/builder/transaction/mod.rs
```diff
@@ -639,6 +646,70 @@ pub fn create_move_to_vault_txhandler(
...
+/// using Sighash Single | AnyoneCanPay. This function will combine the inputs and outputs of the
+/// transactions into a single transaction. Beware, this may be dangerous, as there are no checks.
+pub fn combine_emergency_stop_txhandler(
+    txs: Vec<(Txid, Transaction)>,
+    add_anchor: bool,
+) -> Transaction {
+    // Sanity checks
+    for (_, tx) in &txs {
+        if tx.input.len() != 1 {
+            panic!("Expected only one input per transaction");
```
mmtftr, line 689: remove panic

core/src/builder/transaction/mod.rs
```diff
@@ -639,6 +646,70 @@ pub fn create_move_to_vault_txhandler(
...
+        }
+        if tx.output.len() != 1 {
+            panic!("Expected only one output per transaction");
```
mmtftr, line 692: remove panic

Cargo.toml
```diff
@@ -3,8 +3,10 @@ resolver = "2"
 members = [
   "core",
   "circuits-lib",
-  "bridge-circuit-host"
-  ]
+  # "risc0-circuits/bridge-circuit",
```
mmtftr, line 6: why these comments?

core/src/database/aggregator.rs
```diff
@@ -0,0 +1,146 @@
...
+        let query = query_builder.build_query_as::<(TxidDB, Vec<u8>)>();
+
+        let results: Vec<(TxidDB, Vec<u8>)> =
+            execute_query_with_tx!(self.connection, tx, query, fetch_all)?;
+
+        Ok(results
+            .into_iter()
+            .map(|(txid, tx_data)| {
+                let tx = consensus::deserialize(&tx_data)
+                    .expect("Failed to deserialize emergency stop tx");
```
mmtftr, line 59: should be returned as an error, not a panic

core/src/rpc/aggregator.rs
```diff
@@ -1080,6 +1231,29 @@ impl ClementineAggregator for Aggregator {
...
+
+    async fn internal_create_emergency_stop_tx(
+        &self,
+        request: Request<clementine::CreateEmergencyStopTxRequest>,
+    ) -> Result<Response<clementine::SignedTxWithType>, Status> {
+        let inner_request = request.into_inner();
+        let txids: Vec<Txid> = inner_request
+            .txids
+            .into_iter()
+            .map(|txid| Txid::from_slice(&txid.txid).expect("Failed to parse txid"))
```
mmtftr, line 1243: properly return error instead of panic

mmtftr (changes requested):
Good to merge after removing all possibly panicking code

mmtftr (approved):

ceyhunsen (approved):

PR #739, opened 2025-05-05
Codecov support

ceyhunsen:
# Description

Adds Codecov CI/CD job.

- https://about.codecov.io/
- https://docs.codecov.com/docs/quick-start


ceyhunsen:
This won't run on main branch. Will that be a problem?

mmtftr (commented):
Thanks, can we have some docs for this in the README?

ceyhunsen:
What kind of docs?

mmtftr:
Section on CI/test

we generate codecov reports automatically in CI, you can find these at X

ceyhunsen:
Well, the thing is, I don't know how will they get presented to us. All I know is Codecov will make a comment on PRs, like this https://github.com/chainwayxyz/citrea/pull/2310#issuecomment-2846827038. So, if there will be a thing to show, we are going to discover later.

ceyhunsen:
I was thinking maybe I can also mark coverage job as `continue-on-error` because it is very flaky. If we have reports let's say every 5 PR, should be acceptable until flakiness is reduced.

.github/workflows/build_and_test.yml
```diff
@@ -145,3 +145,43 @@ jobs:
...
+      - uses: ./.github/actions/build-prerequisites
+      - uses: ./.github/actions/test-prerequisites
+      - uses: taiki-e/install-action@cargo-llvm-cov
+
+      - name: Run coverage
+        run: cargo llvm-cov --locked --lcov --output-path lcov.info
+      - name: Upload coverage
+        uses: codecov/codecov-action@v4
+        with:
+          fail_ci_if_error: true # optional (default = false)
```
mmtftr, line 185: disable this please, we don't want to be blocked by codecov's services being down or similar

mmtftr (approved):

PR #740, opened 2025-05-05
Remove Kickoff Tx from Bridge Circuit

aoengin:


ozankaymak (commented):
LGTM. Docs need update

PR #744, opened 2025-05-06
Fix Journal Mismatch

aoengin:


atacann:
Does it not make sense to use SuccinctBridgeCircuitPublicInputs to create journal hash inside bridge_circuit() too. Also while looking at the bridge_circuit, shouldn't it also check if given hcp's last blockhash is equal to latest_blockhash.

Also in bridge_circuit()

    assert_eq!(
        user_wd_txid, input.payout_spv.transaction.input[0].previous_output.txid,
        "Invalid withdrawal transaction ID"
    );

What does this line check? Does it only check withdrawal utxo's txid? Shouldn't it fully check the utxo?

ozankaymak (approved):

PR #749, opened 2025-05-12
Emergency stop CLI Endpoint

ekrembal:


core/src/builder/transaction/mod.rs
```diff
@@ -659,7 +665,7 @@ pub fn create_emergency_stop_txhandler(
             DEFAULT_SEQUENCE,
         )
         .add_output(UnspentTxOut::from_scripts(
-            paramset.bridge_amount - ANCHOR_AMOUNT,
+            paramset.bridge_amount - ANCHOR_AMOUNT.checked_mul(3).unwrap(),
```
mmtftr, line 668: ANCHOR_AMOUNT * 3 is better since this'll never fail

core/src/builder/transaction/mod.rs
```diff
@@ -669,6 +675,34 @@ pub fn create_emergency_stop_txhandler(
     Ok(builder)
 }
 
+/// We assume that the vector of (Txid, Transaction) includes input-output pairs which are signed
+/// using Sighash Single | AnyoneCanPay. This function will combine the inputs and outputs of the
+/// transactions into a single transaction. Beware, this may be dangerous, as there are no checks.
+pub fn combine_emergency_stop_txhandler(txs: Vec<(Txid, Transaction)>) -> Transaction {
+    for (_, tx) in &txs {
+        if tx.input.len() != 1 {
+            panic!("Expected only one input per transaction");
```
mmtftr, line 684: no panics, we don't want to crash the aggregator server when combining txs.

core/src/rpc/clementine.proto
```diff
@@ -565,4 +569,6 @@ service ClementineAggregator {
   rpc InternalSendTx(SendTxRequest) returns (Empty) {}
 
   rpc GetNofnAggregatedXonlyPk(Empty) returns (NofnResponse) {}
+
+  rpc InternalSendEmergencyStopTx(TxidList) returns (Txid) {}
```
mmtftr, line 573: docs here: this is a list of movetx txids

ekrembal: 0dfd599

core/src/builder/transaction/mod.rs
ekrembal: 3d3f30e

Fixed to not use ANCHOR_AMOUNT

core/src/builder/transaction/mod.rs
ekrembal: it's a sanity check. ok for now

PR #752, opened 2025-05-13
Update citrea

ekrembal:
https://github.com/chainwayxyz/citrea/pull/2290


ceyhunsen (dismissed):
didnt notice suggested changes are in tests. can be ignored for fast merge

mmtftr (dismissed):
Looks good to me after ceyhun's comments

core/src/citrea/mod.rs
```diff
@@ -485,111 +483,3 @@ type CitreaContract = BRIDGE_CONTRACT::BRIDGE_CONTRACTInstance<
...
-            println!("Logs from direct call: {:?}", logs_from_direct_call);
-            assert!(logs_from_direct_call.is_err());
-
-            Ok(())
-        }
-    }
-
-    #[tokio::test]
-    #[ignore = "Includes code that won't change much and the test itself is too flaky; Ignoring..."]
-    async fn citrea_get_logs_limit_check() -> citrea_e2e::Result<()> {
```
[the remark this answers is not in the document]
ekrembal: It's not a test we need to have, it's about Citrea mostly

core/src/test/common/citrea/parameters.rs
```diff
@@ -120,22 +110,173 @@ pub fn get_transaction_params(
...
+    )?;
+
+    let txin = builder::transaction::input::SpendableTxIn::new(
+        withdrawal_dust_utxo.outpoint,
+        withdrawal_dust_utxo.txout.clone(),
+        vec![],
+        None,
+    );
+
+    let undpent_txout =
```
mmtftr, line 236: typo
`unspent_txout`

core/src/test/common/citrea/parameters.rs
```diff
@@ -120,22 +110,173 @@ pub fn get_transaction_params(
...
+        builder::transaction::output::UnspentTxOut::from_partial(payout_output.clone());
+
+    let mut tx = builder::transaction::TxHandlerBuilder::new(TransactionType::Payout)
+        .add_input(
+            NormalSignatureKind::NotStored,
+            txin,
+            SpendPath::KeySpend,
+            builder::transaction::DEFAULT_SEQUENCE,
+        )
+        .add_output(undpent_txout.clone())
```
mmtftr, line 246: is this correct? there is no op_return here? (if we're not going to need op_return for some reason, can we add some explanation?)

also we can use `create_payout_txhandler` instead if the op_return will be here.

ekrembal: Yes, we can't put the op_return here since this is kind of psbt, it is not the full payout tx

core/src/test/common/mod.rs
```diff
@@ -359,34 +358,39 @@ pub async fn run_single_deposit<C: CitreaClientT>(
...
+    //     transaction.clone(),
+    //     block,
+    //     block_height as u32,
+    //     move_txid,
+    // ).await?;
+    // println!("Move tx Transaction params: {:?}", transaction_params);
+    // println!(
+    //     "Move tx: {:?}",
+    //     hex::encode(bitcoin::consensus::serialize(&transaction))
+    // );
```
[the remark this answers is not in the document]
ekrembal: I sometimes need this raw move tx, but it is not part of the test
I use it to create test cases for the Bridge conract, let's leave it as it is

mmtftr (approved):

PR #753, opened 2025-05-14
Move citrea mock to `test/` and make citrea module a single file

ceyhunsen:
# Description

Move citrea mock to test/ and make citrea module a single file.

With this PR:

- Citrea client is located at `core/src/citrea.rs`
- Mock client is located at `core/src/test/common/citrea/client_mock.rs`
- Bridge JSON is located at `scripts/Bridge.json`


ekrembal:
Let's merge this after #752

PR #756, opened 2025-05-14
use correct domain name for tls config

ekrembal:


core/src/rpc/mod.rs
```diff
@@ -89,29 +99,33 @@ where
...
+                    .split("://")
+                    .nth(1)
+                    .unwrap_or("localhost")
+                    .split(':')
+                    .next()
+                    .unwrap_or("localhost");
+                let tls_config = if use_client_cert {
+                    let client_identity = Identity::from_pem(client_cert, client_key);
+                    ClientTlsConfig::new()
+                        .domain_name(domain_name)
```
mmtftr, line 120: remove domain_name as discussed

mmtftr (changes requested):

aoengin (approved):

PR #758, opened 2025-05-15
Fix SPV by implementing mid-state hashes

### LINKED ISSUE #506: Fix SPV proof verification in risc0

ozankaymak:
# Description
Fixes SPV by using first-digests instead of the nodes directly.
Fixes #506


ekrembal (approved):
LGMT, More docs can be added

aoengin (approved):

PR #762, opened 2025-05-16
Fix clippy because new lints

ceyhunsen:
# Description

- Wraps some of the `Status` errors with `Box`
- Allows some of the `clippy::result_large_err`
- Converts some of the ?s to `unwrap`s in tests because no need to add additional conversion logic


mmtftr (approved):
Looks good to me

PR #764, opened 2025-05-19
Optimistic payout

atacann:
Adds optimistic payout. All verifiers that signed the deposit directly sign a tx that pays the pegout to user directly from movetx.

### DIFF
--- a/core/src/database/verifier.rs
+++ b/core/src/database/verifier.rs
@@ -68,17 +68,15 @@ impl Database {
         tx: Option<DatabaseTransaction<'_, '_>>,
         citrea_idx: u32,
     ) -> Result<Option<Txid>, BridgeError> {
-        let query = sqlx::query_as::<_, (Option<TxidDB>,)>(
+        let query = sqlx::query_as::<_, (TxidDB,)>(
             "SELECT move_to_vault_txid FROM withdrawals WHERE idx = $1",
         )
         .bind(i32::try_from(citrea_idx).wrap_err("Failed to convert citrea index to i32")?);
 
-        let result: Option<(Option<TxidDB>,)> =
+        let result: Option<(TxidDB,)> =
             execute_query_with_tx!(self.connection, tx, query, fetch_optional)?;
 
-        Ok(result.map(|(move_to_vault_txid,)| {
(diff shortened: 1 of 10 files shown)


core/src/database/verifier.rs
```diff
@@ -63,6 +63,22 @@ impl Database {
         Ok(())
     }
 
+    pub async fn get_move_to_vault_txid_from_citrea_deposit(
```
ceyhunsen, line 66: we can now have test for this

core/src/verifier.rs
```diff
@@ -826,6 +826,103 @@ where
...
+        nonce_session_id: u32,
+        agg_nonce: MusigAggNonce,
+        withdrawal_id: u32,
+        input_signature: Signature,
+        input_outpoint: OutPoint,
+        output_script_pubkey: ScriptBuf,
+        output_amount: Amount,
+    ) -> Result<MusigPartialSignature, BridgeError> {
+        // check if withdrawal is valid first
+        let withdrawal = self
```
ceyhunsen, line 840: dont you looking at deposits here?

ceyhunsen (changes requested):

ceyhunsen (dismissed):

core/src/rpc/aggregator.rs
```diff
@@ -810,6 +825,141 @@ impl ClementineAggregator for Aggregator {
...
+                .await
+                .map_to_status()
+                .map_err(|e| *e)?;
+            let withdrawal_utxo = UTXO {
+                outpoint: input_outpoint,
+                txout: withdrawal_prevout,
+            };
+            let output_txout = TxOut {
+                value: output_amount,
+                script_pubkey: output_script_pubkey,
```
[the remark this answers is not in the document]
mmtftr: input_signature only works for the correct output_txout 

verifier checks input_outpoint before signing:
https://github.com/chainwayxyz/clementine/pull/764/files/65525fd950b3d77302ac22206ed7b7f35bd0cf42#diff-74c4ef62335a6266f398dcca891b1dfa3e2456a3bd924e36f197846526bb67a2R857

mmtftr (approved):

PR #769, opened 2025-05-19
Scripts dir cleanup

### LINKED ISSUE #766: Move important files inside scripts/ to elsewhere
I think scripts folder should only include helper scripts like run.sh and generate_certs.sh

ceyhunsen:
# Description

- Deleted some of the old scripts
- Moved db sql files to `database` module
- Moved test config file to `config` module because protocol paramset config is also there
  - Also renamed
- Deleted old hcp bins
- Updated readme
- Moved `core/tests` into `core/src/test` because that dir was triggering `Cargo` to check for integration tests

## Linked Issues

- Closes #766 

## Testing

Only ran `cargo test`


mmtftr (commented):
Thanks, can we also add some docs on existing `scripts/` directory files? What are they used for? When are they used? etc.

ceyhunsen:
Added small section about script directory @mmtftr. That directory should be self-explanatory. That's why I didn't write anything specific about each script.

Issue #765, opened 2025-05-19
Don't send round1 tx initially


ekrembal:
- [x] Operator reimbursment wallet address should be a config param, if it doesn't exist we can do getNewAddress
- [x] Operator collateral funding outpoint should be a config param, if it doesn't exist we can do sendToAddress
- [ ] round1 tx should't be added to tx_sender in the operator_get_params

PR #772, opened 2025-05-19
Add feature flag for automations (State Machine, TxSender)

### LINKED ISSUE #768: Feature flag state manager & tx sender

mmtftr:
# Description

Put the state machine and the TxSender behind feature flags.

Currently:
1 - BlockCache and ENTITY_TYPE have been moved out into `utils` and `builder` modules.
2 - All state-manager related impls and types have been moved out behind feature flag directives.
3 - Add CI job that checks compilation without state machine
4 - Refactor Aggregator `new_deposit`: no longer sends `move_to_vault_tx`, instead returns raw tx, which can be sent using `send_move_to_vault_tx` with the appropriate deposit params.
5 - Move TxSender task/calls behind a feature flag. (State machine will likely depend on the TxSender feature, as it doesn't make sense otherwise)
6 - Verifier now independently fetches and handles finalized blocks when automation is disabled.
7 - New deposit test script, also included run.sh in the repo.

## Caveats

- All tx-sending features are disabled without automation. This means that operator cannot get reimbursed automatically now. We need to implement another tool/system to facilitate reimbursing.
- Tests don't work without automation, so non-automation build of the bridge is not tested at all.
- Aggregator does not send the movetx on deposit anymore. The movetx must be manually sent using the `send_move_to_vault_tx` (BREAKING CHANGE for clementine-backend)
- #799

## Linked Issues

- Closes #768 
- Depends on #764 

## Testing

Currently, tests run only with state machines and TxSender ON. We may want to have non-automation tests.


ozankaymak (commented):
While you are on this, can you add documentation for the state machine and tx sender? It think they are sort of hard to understand; therefore, hard to use

core/src/operator.rs
```diff
@@ -310,27 +347,9 @@ where
...
-                first_round_tx.get_cached_tx(),
-                FeePayingType::CPFP,
-                None,
-                &[],
-                &[],
-                &[],
-                &[],
-            )
+        #[cfg(feature = "automation")]
+        self.send_initial_round_tx(first_round_tx.get_cached_tx())
```
mmtftr, line 351: https://github.com/chainwayxyz/clementine/issues/765

core/src/rpc/aggregator.rs
```diff
@@ -1527,109 +1567,24 @@ mod tests {
...
-        )
-        .await
-        .wrap_err_with(|| eyre::eyre!("MoveTx did not land onchain"))
-        .unwrap();
-
-        assert!(tx.confirmations.unwrap() > 0);
-    }
-
-    #[tokio::test(flavor = "multi_thread")]
-    async fn aggregator_double_deposit() {
```
mmtftr, line 1568: it seems a test was removed for some reason, put it back

core/src/rpc/operator.rs
```diff
@@ -296,10 +296,20 @@ where
...
+
+            self.operator.end_round(&mut dbtx).await?;
+
+            dbtx.commit().await.expect("Failed to commit transaction");
+            Ok(Response::new(Empty {}))
+        }
+
+        #[cfg(not(feature = "automation"))]
+        Err(Status::unimplemented(
+            "This method is only available in automation mode",
```
mmtftr, line 311: Automation is not enabled. Operator does not manage its rounds txs.

core/src/test/common/mod.rs
```diff
@@ -489,6 +508,7 @@ fn sign_nofn_deposit_tx(
     tx
 }
 
+#[cfg(feature = "automation")]
 pub async fn run_replacement_deposit(
```
mmtftr, line 512: this can potentially be refactored to not require automation.

core/src/verifier.rs
```diff
@@ -1246,6 +1261,7 @@ where
                 | TransactionType::KickoffNotFinalized
                 | TransactionType::LatestBlockhashTimeout
                 | TransactionType::OperatorChallengeNack(_) => {
+                    #[cfg(feature = "automation")]
```
mmtftr, line 1264: we need to add another thing here for when a malicious kickoff is detected

core/src/verifier.rs
```diff
@@ -1344,30 +1360,34 @@ where
...
-        let mut dbtx = self.db.begin_transaction().await?;
-        self.tx_sender
-            .add_tx_to_queue(
-                &mut dbtx,
-                tx_type,
-                &challenge_tx,
-                &[],
-                Some(TxMetadata {
+        #[cfg(feature = "automation")]
+        {
```
mmtftr, line 1364: we need another thing here for when we need to send WT challegnes

core/src/rpc/aggregator.rs
```diff
@@ -604,39 +604,8 @@ impl Aggregator {
...
-        dbtx.commit()
-            .await
-            .map_err(|e| Status::internal(format!("Failed to commit db transaction: {}", e)))?;
-
-        tracing::info!(
-            "move_tx: {}",
-            hex::encode(bitcoin::consensus::serialize(&move_tx))
-        );
 
+        // TODO: Sign the transaction correctly after we create taproot witness generation functions
```
mmtftr, line 608: why is there a todo like this here? rm

mmtftr: done

core/src/rpc/aggregator.rs
```diff
@@ -1359,11 +1339,14 @@ impl ClementineAggregator for Aggregator {
...
+            .create_movetx(move_to_vault_sigs, movetx_agg_nonce, deposit_params)
             .await?;
-        let txid = *signed_movetx_handler.get_txid();
 
-        Ok(Response::new(txid.into()))
+        let raw_signed_tx = RawSignedTx {
+            raw_tx: bitcoin::consensus::serialize(&signed_movetx_handler.get_cached_tx()),
+        };
+
+        Ok(Response::new(raw_signed_tx))
```
atacann, line 1349: I think it would be better to send movetx here if automation is enabled for conveniency, but returning movetx instead of txid is good in case its not enabled.

mmtftr: I think it's better to have a consistent interface like this. In this way, if we incorrectly compile the aggregator, we'll get a clear error mesage. Otherwise, we would get a silent failure which we'd have to debug (i.e. we'd be debugging why movetx wasn't being sent)

core/src/rpc/aggregator.rs
mmtftr: now they're actually in reverse order, put them in the right order

core/src/verifier.rs
mmtftr: https://github.com/chainwayxyz/clementine/issues/799

mmtftr:
- Check if `handle_finalized_block` is idempotent. We need this so that state manager can be enabled after deployment

core/src/utils.rs
```diff
@@ -376,3 +376,14 @@ pub struct RbfSigningInfo {
     pub vout: u32,
     pub tweak_merkle_root: Option<TapNodeHash>,
 }
+pub trait Last20Bytes {
```
ceyhunsen, line 379: Where is this used and why it isn't just a function?

scripts/run.sh
ceyhunsen, line 1: There is `run-test.sh`. If you want to make changes, please do that in that file.

ceyhunsen (changes requested):

core/src/utils.rs
mmtftr: this was merged from master, you're seeing it as new because you're looking at commit differences, resolving

scripts/run.sh
mmtftr: done

ceyhunsen (approved):

ceyhunsen (approved):

PR #775, opened 2025-05-20
Implement annex support for watchtower challenge tx in bridge circuit

ozankaymak:
# Description
Implement annex support for the bridge circuit with a test using a mainnet transaction with annex


aoengin (approved):
LGTM!

ozankaymak:
I see no difference in terms of the Witness being 4 mb (say, a watchtower is inscribing something) and using an annex.

PR #776, opened 2025-05-20
Extended RPC unit tests and documentation

ceyhunsen:
# Description

For extended RPC module:

- Adds unit tests
- Adds documentation


core/src/extended_rpc.rs
```diff
@@ -1,9 +1,22 @@
...
-
-use std::str::FromStr;
-use std::sync::Arc;
+//! Extended RPC interface communicates with the Bitcoin node. It features some
+//! common wrappers around typical RPC operations as well as direct
+//! communication interface with the node.
+//!
+//! ## Tests
+//!
+//! On tests, a new Bitcoin node can be created using [`crate::test::common::create_regtest_rpc`].
```
mmtftr, line 9: the returned value is a handle to the process that's managed by the test. dropping the handle will kill the process.

also not sure if this is the right place to put this. I'd rather put this section inside the common test utils module where this is defined.

ceyhunsen: I wanted this to be useful for newcomers. But yeah, it shouldn't be here. So, I redirected users to that module. This PR is out of scope for that documentation but I will keep in mind your comments. Once I jump there, I will be adding these. Changing commit: https://github.com/chainwayxyz/clementine/pull/776/commits/525fa4435a4d836e142eb97909827f834e933315

core/src/extended_rpc.rs
```diff
@@ -152,8 +196,25 @@ impl ExtendedRpc {
...
+    /// # Returns
+    ///
+    /// - [`bool`]: `true` if the UTXO is spent, `false` otherwise.
+    ///
+    /// # Errors
+    ///
+    /// - [`BitcoinRPCError`]: If the transaction is not confirmed or if there
+    ///   was an error retrieving the transaction output.
     pub async fn is_utxo_spent(&self, outpoint: &OutPoint) -> Result<bool> {
+        if !self.is_tx_on_chain(&outpoint.txid).await? {
```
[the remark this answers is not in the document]
ceyhunsen: Already has that

core/src/extended_rpc.rs
```diff
@@ -1,9 +1,14 @@
...
-
-use std::str::FromStr;
-use std::sync::Arc;
+//! Extended RPC interface communicates with the Bitcoin node. It features some
+//! common wrappers around typical RPC operations as well as direct
+//! communication interface with the Bitcoin node.
+//!
+//! ## Tests
+//!
+//! On tests, Bitcoind node and client are usually created using
```
mmtftr, line 9: nit: In tests

mmtftr (dismissed):

core/src/extended_rpc.rs
ceyhunsen: https://github.com/chainwayxyz/clementine/pull/776/commits/29e14e2fef335a6519e86492f33e4c0048be9a67

atacann (dismissed):
LGTM, there are some comments unrelated to this PR.

According to Erdem if a tx is in a non-canonical block the rpc will still return the blockhash of that block. So is_tx_on_chain might return true for tx's in reorged blocks.

Also the places where we do tx.output[vout] can panic with wrong vout.

atacann (approved):

PR #797, opened 2025-05-27
Confirm fee payer UTXOs before trying to send tx in CPFP

ceyhunsen:
# Description

Simplifies testing steps and reduces flakiness.

- Adds `confirm_fee_payer_utxos`
- Deletes some repeated code to use common test utilities

## Linked Issues

None

### DIFF
--- a/core/src/database/test.rs
+++ b/core/src/database/test.rs
@@ -13,8 +13,8 @@ impl Database {
         &self,
         tx: Option<DatabaseTransaction<'_, '_>>,
         tx_id: u32,
-    ) -> Result<Vec<(Txid, u32, Amount, bool)>, BridgeError> {
-        let query = sqlx::query_as::<_, (TxidDB, i32, i64, Option<i32>)>(
+    ) -> Result<Vec<(Txid, u32, Amount)>, BridgeError> {
+        let query = sqlx::query_as::<_, (TxidDB, i32, i64)>(
             r#"
             SELECT fee_payer_txid, vout, amount
             FROM tx_sender_fee_payer_utxos
@@ -23,19 +23,18 @@ impl Database {
         )
         .bind(i32::try_from(tx_id).wrap_err("Failed to convert tx_id to i32")?);
 
-        let results: Vec<(TxidDB, i32, i64, Option<i32>)> =
+        let results: Vec<(TxidDB, i32, i64)> =
             execute_query_with_tx!(self.connection, tx, query, fetch_all)?;
 
         results
             .iter()
-            .map(|(fee_payer_txid, vout, amount, confirmed)| {
+            .map(|(fee_payer_txid, vout, amount)| {
                 Ok((
                     fee_payer_txid.0,
                     u32::try_from(*vout).wrap_err("Failed to convert vout to u32")?,
                     Amount::from_sat(
                         u64::try_from(*amount).wrap_err("Failed to convert amount to u64")?,
                     ),
-                    confirmed.is_some(),
                 ))
             })
             .collect::<Result<Vec<_>, BridgeError>>()


core/src/test/common/mod.rs
```diff
@@ -386,42 +431,9 @@ pub async fn run_single_deposit<C: CitreaClientT>(
...
-    //     block,
-    //     block_height as u32,
-    //     move_txid,
-    // ).await?;
-    // println!("Move tx Transaction params: {:?}", transaction_params);
-    // println!(
-    //     "Move tx: {:?}",
-    //     hex::encode(bitcoin::consensus::serialize(&transaction))
-    // );
-
```
mmtftr, line 424: can we keep this for debugging purposes in the future? I think ekrem wanted to keep it for that purpose, you can add a small comment at the top like `Uncomment to debug teh move tx`

core/src/test/common/tx_utils.rs
```diff
@@ -350,3 +348,60 @@ pub async fn create_tx_sender(
...
+        None,
+    )
+    .await?;
+
+    Ok(())
+}
+
+/// Finds fee payer utxos for a transaction and confirms them.
+#[cfg(feature = "automation")]
+pub async fn confirm_fee_payer_utxos(rpc: &ExtendedRpc, db: Database, txid: Txid) -> Result<()> {
```
mmtftr, line 400: `confirm` is a misleading name. I don't think this should be a function either, it's just two lines, let's just use those two lines.

core/src/test/common/mod.rs
```diff
@@ -753,3 +720,100 @@ pub fn ensure_test_certificates() -> Result<(), std::io::Error> {
...
+        actor::Actor,
+        extended_rpc::ExtendedRpc,
+        test::common::{
+            citrea::MockCitreaClient, create_regtest_rpc, create_test_config_with_thread_name,
+        },
+    };
+    use bitcoin::{absolute::LockTime, transaction::Version, Transaction};
+    use bitcoincore_rpc::RpcApi;
+
+    #[ignore = "Tested everywhere, no need to run again"]
```
mmtftr, line 735: let's just remove it then

core/src/test/common/mod.rs
```diff
@@ -753,3 +720,100 @@ pub fn ensure_test_certificates() -> Result<(), std::io::Error> {
...
+        let regtest = create_regtest_rpc(&mut config).await;
+        let rpc = regtest.rpc().clone();
+        let _ = super::run_single_deposit::<MockCitreaClient>(&mut config, rpc, None)
+            .await
+            .unwrap();
+    }
+
+    #[cfg(feature = "integration_tests")]
+    #[tokio::test]
+    async fn mine_once_after_in_mempool() {
```
mmtftr, line 749: not sure if it makes sense to test the test utilities.

core/src/database/tx_sender.rs
```diff
@@ -831,6 +831,59 @@ impl Database {
         Ok(())
     }
 
+    #[cfg(test)]
+    pub async fn get_fee_payer_utxos_for_tx(
```
mmtftr, line 835: i would rather these impls be inside the `test::common` module to not clutter the production code.

mmtftr (changes requested):

core/src/test/common/mod.rs
ceyhunsen: https://github.com/chainwayxyz/clementine/pull/797/commits/1fee1ac046a25c2b44e0548afb1bbec870973fe8

core/src/test/common/mod.rs
ceyhunsen: https://github.com/chainwayxyz/clementine/pull/797/commits/34836546047b35865ecc6bc3d07cc4c10ee609f4

I thought these helpers are big and would be nice to test their behavior too

core/src/test/common/mod.rs
ceyhunsen: https://github.com/chainwayxyz/clementine/pull/797/commits/34836546047b35865ecc6bc3d07cc4c10ee609f4

I thought these helpers are big and would be nice to test their behavior too

core/src/database/tx_sender.rs
mmtftr: yeah, you need to do it inside `impl Database { }` (you can add impls to the same struct anywhere inside a crate so this is no problem)

PR #798, opened 2025-05-27
Remove secret keys from bridge config and move it to test params

### LINKED ISSUE #609: Secret keys in config file will cause problems
### Expected Behavior

In a non-test environment, secret keys in config file must be omitted.

### Current Behavior

Secret keys in config file in a non-test environment will be read.

### Stps to Reproduce

1. Change public keys in config file
2. Have wrong secret keys in config file

### Current Behavior

_No response_

### Detailed Description

If a user updates public key fields with correct values and forgets to delete optional secret keys fields, an unrelated error to the user will be printed out. This issue will be irrelevant in the future if both public and secret keys fields in the config file is deleted.

ceyhunsen:
# Description

Removes secret key fields from `BridgeConfig` and moves them to test only field.

## Linked Issues

- Closes #609

### DIFF
--- a/core/src/builder/transaction/creator.rs
+++ b/core/src/builder/transaction/creator.rs
@@ -917,7 +917,7 @@ mod tests {
         txs_operator_can_sign
             .extend((0..verifiers.len()).map(TransactionType::WatchtowerChallengeTimeout));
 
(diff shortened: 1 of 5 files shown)


mmtftr (approved):
nice

PR #801, opened 2025-05-29
Enable nonstandard 

atacann:
Enables 0 sat kickoff utxo's and 0 sat anchor outputs. 

New paramset parameter to support this: BRIDGE_NONSTANDARD=true (or false)
kickoff_amount is still seperate, needs to be set to 0 if nonstandard is enabled. (It will still work but the whole point is to make kickoff_amount 0) kickoff_amount can be removed later and standard and nonstandard kickoff_amounts hardcoded in. 

We cant use 0 sat anchors always even if it is standard with bitcoin core 29 because it requires tx to have 0 fee.
Move tx always uses 0 sat anchors even with standard option, this way the deposited amount in movetx is always exactly the bridge amount.

A potential problem: Right now default protocol paramset is set to standard with 55000 sat kickoff_amount, so tests will not test the nonstandard version simultaneously. This PR only changes code so that switching to nonstandard later is easier.

### DIFF
--- a/core/src/builder/transaction/challenge.rs
+++ b/core/src/builder/transaction/challenge.rs
@@ -319,7 +319,7 @@ pub fn create_disprove_txhandler(
 /// This transaction is used to reimburse an operator for a valid challenge, intended to cover their costs for sending asserts transactions,
 /// and potentially cover their opportunity cost as their reimbursements are delayed due to the challenge. This cost of a challenge is also
 /// used to disincentivize sending challenges for kickoffs that are correct. In case the challenge is correct and operator is proved to be
-/// malicious, the challenge cost will be reimbursed using the operator's collateral thats locked in Citrea.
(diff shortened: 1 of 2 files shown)


core/src/tx_sender/mod.rs
```diff
@@ -268,8 +269,14 @@ impl TxSender {
     }
 
     fn is_p2a_anchor(&self, output: &TxOut) -> bool {
-        output.value == builder::transaction::anchor_output().value
-            && output.script_pubkey == builder::transaction::anchor_output().script_pubkey
+        (output.value
+            == builder::transaction::anchor_output(self.paramset.default_anchor_amount()).value
```
mmtftr, line 273:
```suggestion
            == self.paramset.default_anchor_amount()
```

ekrembal (dismissed):

PR #812, opened 2025-06-03
Use ubuntu-latest instead of ubicloud when it is logical to do so

ceyhunsen:
# Description

Converts some of the `runs-on`s to `ubuntu-latest` from `ubicloud-*`.


atacann:
Additional disprove tests are also single threaded I think, they do not need to run on 30 core machines

PR #813, opened 2025-06-03
Disprove TX

aoengin:


ozankaymak (approved):

core/src/test/bitvm_disprove_scripts.rs
```diff
@@ -0,0 +1,739 @@
...
+    }
+
+    async fn run_test(&mut self, f: &mut TestFramework) -> Result<()> {
+        match self.variant {
+            DisproveTestVariant::HealthyState => {
+                tracing::info!("Running healthy state test");
+                self.healthy_state_test(f).await?;
+            }
+            DisproveTestVariant::CorruptedAssert => {
+                tracing::info!("Running disrupted latest block hash commit test");
```
ozankaymak, line 678: This should be something like disrupted operator BitVM assert. not latest block hash

ozankaymak (commented):
And some refactor maybe, but I think it is a topic for another PR

ozankaymak (approved):

core/src/actor.rs
```diff
@@ -579,6 +579,12 @@ impl Actor {
                 .get_deposit_sig_owner()
                 .map(|s| s.sighash_type())?
                 .unwrap_or(TapSighashType::Default);
+
+            if tx_type == TransactionType::DisproveTimeout {
+                let x = spt.get_spend_path();
+                println!("DisproveTimeout: {:?} {:?}", x, spt.get_signature_id());
```
ceyhunsen, line 585: use tracing please

core/src/bitvm_client.rs
```diff
@@ -51,7 +52,11 @@ lazy_static::lazy_static! {
         let start = Instant::now();
 
         let bitvm_cache = {
-            let cache_path = std::env::var("BITVM_CACHE_PATH").unwrap_or_else(|_| "bitvm_cache.bin".to_string());
+            let cache_path = if cfg!(test) && is_dev_mode() {
```
ceyhunsen, line 55: maybe we can add a third check:

```
std::env::var("BITVM_CACHE_PATH").is_err()
```

core/src/builder/address.rs
```diff
@@ -53,6 +54,29 @@ pub fn taproot_builder_with_scripts(scripts: &[ScriptBuf]) -> TaprootBuilder {
     })
 }
 
+pub fn taproot_leaf_depths(scripts: &[ScriptBuf]) -> Vec<(u8, ScriptBuf)> {
```
ceyhunsen, line 57: a test for this would be nice

core/src/builder/transaction/mod.rs
```diff
@@ -575,24 +576,46 @@ pub fn create_replacement_deposit_txhandler(
 pub fn create_disprove_taproot_output(
     operator_timeout_script: Arc<dyn SpendableScript>,
     additional_script: ScriptBuf,
-    disprove_root_hash: HiddenNode,
+    disprove_path: DisprovePath, // <--- use the enum here
```
ceyhunsen, line 579: this is enum, remove comment or add todo to the beginning if there is another problem with it

core/src/builder/transaction/sign.rs
```diff
@@ -173,6 +178,13 @@ pub async fn create_and_sign_txs(
 
         let checked_txhandler = txhandler.promote();
 
+        if tx_type == TransactionType::DisproveTimeout {
+            println!(
```
ceyhunsen, line 182: tracing

core/src/builder/transaction/sign.rs
```diff
@@ -187,6 +199,13 @@ pub async fn create_and_sign_txs(
         }
     }
 
+    println!(
```
ceyhunsen, line 202: tracing

ceyhunsen (changes requested):

core/src/bitvm_client.rs
ozankaymak: Do we need this?

ceyhunsen: maybe we don't. this closes the door of using a custom cache for tests but shouldn't be a problem i guess

core/src/builder/transaction/creator.rs
```diff
@@ -422,7 +423,7 @@ impl ContractContext {
 
     /// Contains all necessary context for creating txhandlers for a specific operator, kickoff utxo, and a deposit
     /// Additionally holds signer of an actor that can generate the actual winternitz public keys.
-    pub fn new_context_for_asserts(
```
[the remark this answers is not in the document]
ozankaymak: I believe it is to be able to generated needed keys for winternitz

core/src/verifier.rs
```diff
@@ -1976,6 +1973,270 @@ where
...
+        // third.reverse();
+
+        let first_box = Box::new(first);
+        let second_box = Box::new(second);
+        let third_box = Box::new(third);
+
+        tracing::info!("Boxes created");
+
+        let res = validate_assertions(
+            &get_ark_verifying_key_dev_mode_bridge(),
```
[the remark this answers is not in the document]
ozankaymak: I think we have that for non dev mode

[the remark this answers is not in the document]
atacann: I meant you use dev mode verifying key by default, you should use normal ark verifying key for non tests, if you already do somewhere I did not see it

core/src/verifier.rs
```diff
@@ -1976,6 +1973,270 @@ where
...
+
+        let deposit_outpoint = deposit_data.get_deposit_outpoint();
+        let paramset = self.config.protocol_paramset();
+
+        // let mut all_assert_commits: Vec<Vec<Vec<u8>>> = Vec::new();
+        let mut g16_public_input_commit: Vec<Vec<Vec<u8>>> = vec![vec![vec![]]; 1];
+        let mut num_u256_commits: Vec<Vec<Vec<u8>>> = vec![vec![vec![]]; 14];
+        let mut intermediate_value_commits: Vec<Vec<Vec<u8>>> = vec![vec![vec![]]; 363];
+        tracing::info!("Number of operator asserts: {}", operator_asserts.len());
+        for i in 0..operator_asserts.len() {
```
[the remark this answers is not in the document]
aoengin: and return an error if not all asserts exist?

ozankaymak (approved):

ozankaymak (dismissed):

mmtftr (changes requested):
ok to merge after removing unsafe

core/src/test/additional_disprove_scripts.rs
```diff
@@ -165,7 +165,8 @@ impl AdditionalDisproveTest {
             )
             .await
             .unwrap(),
-            0
+            0,
```
ceyhunsen, line 168: a sleep before this can decrease flakiness

PR #825, opened 2025-06-12
CLI binary improvements

### LINKED ISSUE #817: Improve bin/cli.rs
- [x] Remove code duplication (parsing etc)
- [x] re-Adding get transaction params
- [x] Adding Citrea Deposit functionality
- [x] Adding Citrea safeWithdraw Functionality

ceyhunsen:
# Description

Removes code duplications and adds new commands.

## Linked Issues

- Closes #817

### DIFF
--- a/core/src/bin/cli.rs
+++ b/core/src/bin/cli.rs
@@ -1,14 +1,10 @@
 //! This module defines a command line interface for the RPC client.
 
-use std::path::PathBuf;
-use std::str::FromStr;
-
-use bitcoin::{hashes::Hash, Block, ScriptBuf, Txid};
+use bitcoin::{hashes::Hash, ScriptBuf};
 use clap::{Parser, Subcommand};
 use clementine_core::{
     config::BridgeConfig,
     deposit::SecurityCouncil,
-    errors::BridgeError,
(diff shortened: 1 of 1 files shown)


cli/src/utils/bitcoin_merkle.rs
```diff
@@ -0,0 +1,174 @@
...
+
+pub fn calculate_double_sha256(input: &[u8]) -> [u8; 32] {
+    let mut hasher = Sha256::default();
+    hasher.update(input);
+    let result = hasher.finalize_reset();
+    hasher.update(result);
+    hasher.finalize().into()
+}
+
+impl BitcoinMerkleTree {
```
atacann, line 18: Is it not possible to use BitcoinMerkleTree in circuits-lib

atacann (changes requested):
I think there are still a lot of code duplication, mainly helper fns for citrea that are in tests module in clementine that is duplicated in now cli crate. I think there are two possible solutions: 1 - move these fns out of tests module so cli can import them from clementine core, or create a new crate for these kinds of util fns that clementine-core tests and cli can import.

PR #840, opened 2025-06-16
Refactor watchtower challenge output types for improved flexibility

ozankaymak:


circuits-lib/src/bridge_circuit/mod.rs
```diff
@@ -441,34 +448,53 @@ pub fn total_work_and_watchtower_flags(
...
-            continue;
-        };
-
-        let compressed_g16_proof: [u8; 128] = [&first_output, &second_output, &third_output[0..64]]
-            .concat()
-            .try_into()
-            .expect("Cannot fail");
+        let compressed_g16_proof: [u8; 128];
+        let total_work: [u8; 16];
+
```
aoengin, line 453: Please check the length of the outputs and skip the current wt tx if the output length is not correct

aoengin (changes requested):

PR #846, opened 2025-06-18
CI/CD overhaul

ceyhunsen:
# Description

- Fixes cache issues across build and test jobs
- Builds and tests with both all and no features, not just with a feature
- Bug template issue fixes 
- Makes coverage tests release
  - Hopes to reduce flakiness
  - Some code not working while in debug mode

## Notes

Building and testing with all and no features means that deployment should be OK with no features enabled. On the other hand, all features also should pass but not a necessity. Because new features may not be ready.


atacann (approved):
I think we should just delete coverage job honestly

PR #851, opened 2025-06-19
Fix Test Flakiness

aoengin:


core/src/builder/transaction/creator.rs
```diff
@@ -685,6 +697,9 @@ pub async fn create_txhandlers(
         .ok_or(TxError::IndexOverflow)?
         .clone();
 
+    tracing::warn! {"Payout tx blockhash pk: {:?}",
```
mmtftr, line 700: debug info == tracing::debug

core/src/verifier.rs
```diff
@@ -1878,6 +1878,8 @@ where
                 .into());
             }
             operator_acks_vec[*idx] = Some(pre_image);
+
+            tracing::warn!("Operator ack for idx {}", idx);
```
mmtftr, line 1882: debug info == tracing::debug

core/src/verifier.rs
```diff
@@ -1904,6 +1906,32 @@ where
             payout_blockhash_new.push(element);
         }
 
+        tracing::warn!(
```
mmtftr, line 1909: debug info == tracing::debug
+ all below

mmtftr (commented):
I think we need to debug why a transport error happens as well instead of just retrying. I'm guessing that some kind of resource is exhausted in the system with all of the tests running at the same time

core/src/test/additional_disprove_scripts.rs
```diff
@@ -934,6 +937,8 @@ impl TestCase for AdditionalDisproveTest {
 #[tokio::test]
 #[ignore = "This test is too slow, run separately"]
 async fn additional_disprove_script_test_disrupted_latest_block_hash() -> Result<()> {
+    initialize_logger(Some(::tracing::level_filters::LevelFilter::DEBUG))
```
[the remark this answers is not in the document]
aoengin: Nope, when the IS_CI flag is set to true or 1, the inner logic will override it. Also, if you set RUST_LOG when running it locally, that will override too. This just sets the default log level there are plenty of ways to override it.

core/src/utils.rs
```diff
@@ -35,39 +37,78 @@ use tracing_subscriber::{fmt, EnvFilter, Registry};
...
+                .with_thread_names(true)
+                .with_filter(LevelFilter::INFO)
+                .boxed();
+
+            layers.push(file_layer);
+        }
+
+        let subscriber = Registry::default().with(layers);
+
+        let res = tracing::subscriber::set_global_default(subscriber);
```
mmtftr, line 83: duplicate of below.

can we consolidate by matching logic, then using subfunctions to isolate the exact subscribers. Illustration:

```rs

fn try_set_global_subscriber<S>(subscriber: S) where S: Subscriber + ... {
  // try logic
}
fn initialize_logger() {
  // ... existing code
  if is_ci() {
    try_set_global_subscriber(env_subscriber_with_file())?;
  } else {
    match is_json_logs() {
      true => try_set_global_subscriber(env_subscriber_to_json())?,
      false => try_set_global_subscriber(env_subscriber_to_human())?,
    }
  };
}
```

This will make it much more readable and easier to manage.

bridge-circuit-host/src/bridge_circuit_host.rs
```diff
@@ -175,7 +175,7 @@ pub fn prove_bridge_circuit(
     let circuit_g16_proof = CircuitGroth16Proof::from_seal(risc0_g16_256);
     let ark_groth16_proof: ark_groth16::Proof<Bn254> = circuit_g16_proof.into();
 
-    tracing::warn!(
+    tracing::info!(
```
mmtftr, line 178: Can we make these tracing::debug please? Somebody using the bridge is not going to be interested in seeing this information as this is debug info, meant for us to see.

core/src/builder/transaction/creator.rs
```diff
@@ -697,7 +697,7 @@ pub async fn create_txhandlers(
         .ok_or(TxError::IndexOverflow)?
         .clone();
 
-    tracing::warn! {"Payout tx blockhash pk: {:?}",
+    tracing::info! {"Payout tx blockhash pk: {:?}",
```
mmtftr, line 700: debug please

mmtftr (changes requested):

core/src/utils.rs
atacann: Also written in current way, ci console logs are always set to warn. I think it might be better to remove IS_CI, INFO_LOG_FILE option is enough. And use previous json_layer/standard_layer in addition to file_layer. 

Also can the file get viewed nicely without json? Might be better to change it to json (I saw it has with_ansi(false), should be fine then)

core/src/builder/transaction/creator.rs
```diff
@@ -672,6 +672,18 @@ pub async fn create_txhandlers(
         context.paramset.genesis_chain_state_hash,
     );
 
+    tracing::info!(
```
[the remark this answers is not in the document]
atacann: Yes this log can get printed a lot.

atacann (changes requested):
You didn't set some logs here and the log file to debug because it gets spammed a lot right?
I think some of these logs that have info is fine because they get printed only during a challenge, which will be very very rare. But a better solution can be to fix the debug spam.

DB query logs, signature collection logs, and h2 http logs are getting printed to debug, changing them to trace can reduce a lot of spam (not related to this PR).

.github/workflows/build_and_test.yml
```diff
@@ -12,6 +12,7 @@ env:
   RISC0_DEV_MODE: 1
   ENABLE_HEADER_CHAIN_PROVER: 1
   RUST_MIN_STACK: 33554432
+  IS_CI: 1
```
mmtftr, line 15: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables

Can we use `CI`, it's always set to `true`

core/src/utils.rs
```diff
@@ -35,7 +37,102 @@ use tracing_subscriber::{fmt, EnvFilter, Registry};
...
+
+    let file = File::create(path).map_err(|e| BridgeError::ConfigError(e.to_string()))?;
+
+    let file_filter = EnvFilter::from_default_env()
+        .add_directive("info".parse().expect("It should parse info level"))
+        .add_directive("ci=debug".parse().expect("It should parse ci debug level"));
+
+    let console_filter = EnvFilter::builder()
+        .with_default_directive(LevelFilter::WARN.into())
+        .parse_lossy(""); // Parse an empty string to use the default filter, this prevents overriding the default filter with env vars.
```
mmtftr, line 102: can we parse from env here too for consistency?

mmtftr (commented):
Thanks, small comments

mmtftr (approved):
Thanks for the changes

PR #861, opened 2025-06-20
Atacan/tx sender already funded

atacann:
Add alreadyfunded option to txsender, to send a tx as is


ceyhunsen:
What good does this PR do? Why can't we just use `rpc.send_raw_transaction()` without adding it to `tx_sender`?

mmtftr (changes requested):
naming changes as requested before

ekrembal (approved):

Issue #746, opened 2025-05-06
Re-enable circuit e2e testing with data


ozankaymak:
### Issue Description

Re-enable the circuit testing with produced data in bridge_circuit_host with different scenarios.
The data is generated in PR #864 for Issue #855. This task also includes writing tests using the pre-generated data from the mentioned PR.

PR #864, opened 2025-06-23
[Testing] Guest binary tests, better watchtower tests

### LINKED ISSUE #845: Create method id consts for all circuits and networks and verify them in a test
Create consts for method ids of elfs (circuit-network pair), and in a test compute the method id's of elfs and see if the consts match. 

- Helps having an easy way to check current method ids (bridge circuit method id is needed for config).
- Helps noticing if any method id changes.

### LINKED ISSUE #855: 1MB watchtower challenge tx test
We need to test extreme cases for watchtowers:
1. 1MB tx
2. 4MB tx
3. Wt Challenge with an Annex

### LINKED ISSUE #859: Ensuring disprove type
### Proposal Description

When testing disprove/additional disprove, there is a need to check which path is used onchain to make sure the logic is correct.

ozankaymak:
# Description
This PR introduces the different types of transactions that a verifier can use for sending transactions. Also includes tests for guest binaries for consistency. Also fixes a small bug on annex check.

## Linked Issues

- Closes #859 
- Closes #855 
- Closes #845 

## Testing
Added e2e tests, ignored for now
Also added binary tests

## Docs

have the docs been updated?

### DIFF
--- a/bridge-circuit-host/src/bridge_circuit_host.rs
+++ b/bridge-circuit-host/src/bridge_circuit_host.rs
@@ -1,5 +1,3 @@
-use std::io::Write;
-
 use crate::docker::{stark_to_bitvm2_g16, stark_to_bitvm2_g16_dev_mode};
 use crate::structs::{
     BridgeCircuitBitvmInputs, BridgeCircuitHostParams, SuccinctBridgeCircuitPublicInputs,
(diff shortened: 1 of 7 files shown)


ozankaymak:
Also, see [here](https://github.com/chainwayxyz/clementine/issues/746) for the circuit testing with data.

aoengin (approved):
lgtm

core/src/config/protocol.rs
```diff
@@ -306,4 +321,70 @@ mod tests {
             hex::encode(calculated_bridge_circuit_constant)
         );
     }
+
+    #[test]
+    fn test_header_chain_method_ids() {
```
atacann, line 326: Can you create constants for bridge circuit method ids, and test them here too. Also create constants for whatever is in config and test it too (like in test_calculate_succinct_output_prefix but not only for regtest, for all networks). Basically I want to just copy paste that const so that I can use it in config. I dont want to do it manually. Any mistake in config in deployment really consumes time

atacann (changes requested):

atacann (dismissed):
Thanks

PR #868, opened 2025-06-24
Concurrent deposit and withdraw tests

ceyhunsen:
# Description

Adds concurrent deposits and withdrawals test. Also blocks `deposit_sign` to hold mutex lock for `verifier.nonces` for the signing session. It can take and update at the start and end, respectively.

### DIFF
--- a/core/src/rpc/aggregator.rs
+++ b/core/src/rpc/aggregator.rs
@@ -900,9 +900,9 @@ impl ClementineAggregator for Aggregator {
                 .map_to_status()?;
             let agg_nonce = aggregate_nonces(pub_nonces.iter().collect::<Vec<_>>().as_slice());
             // send the agg nonce to the verifiers to sign the optimistic payout tx
-            let verifier_clients = self.get_verifier_clients();
+            let verifier_clients = verifiers.clients();
             let payout_sigs = verifier_clients
-                .iter()
+                .into_iter()
                 .zip(first_responses)
                 .map(|(client, first_response)| {
                     let mut client = client.clone();


core/src/test/deposit_and_withdraw_e2e.rs
```diff
@@ -1560,3 +1570,249 @@ async fn mock_citrea_run_malicious_after_exit() {
...
+            deposit_outpoint,
+            deposit_type: DepositType::BaseDeposit(BaseDepositData {
+                evm_address,
+                recovery_taproot_address: actor.address.as_unchecked().to_owned(),
+            }),
+        };
+
+        let deposit: Deposit = deposit_info.clone().into();
+
+        let movetx = aggregator
```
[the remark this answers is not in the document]
ceyhunsen: no, not intended. i guess this was an accident while doing a refactor.

core/src/test/deposit_and_withdraw_e2e.rs
```diff
@@ -1552,3 +1555,244 @@ async fn mock_citrea_run_malicious_after_exit() {
 
     assert!(tx.output[0].value != config.protocol_paramset().operator_challenge_amount);
 }
+
+#[tokio::test(flavor = "multi_thread")]
+async fn concurrent_deposits_and_withdrawals() {
```
mmtftr, line 1560: Small comment explaining what the test does would be appreciated since it contains ~300 lines of code.

core/src/verifier.rs
```diff
@@ -2375,20 +2387,20 @@ where
...
             .get_citrea_l2_height_range(block_height.into(), timeout)
-            .await;
-        if let Err(e) = l2_range_result {
-            tracing::error!("Error getting citrea l2 height range: {:?}", e);
-            return Err(e);
-        }
-
-        let (l2_height_start, l2_height_end) =
-            l2_range_result.expect("Failed to get citrea l2 height range");
+            .await
```
mmtftr, line 2393: thanks for removing the except here. Can we write it with inspect_err instead?

mmtftr (dismissed):

core/src/verifier.rs
ceyhunsen: Err needs to be returned. So, inspect_err can't be used here.

ceyhunsen: Yeah, you are right. Forgot to use `?` and tried to return err in closure 🤦. https://github.com/chainwayxyz/clementine/pull/868/commits/49e79f4b9eadd655b052b3aefd7739ea2715cb8f

atacann (changes requested):
Looks good, can you add concurrent optimistic payouts to the test also?

citrea-tsc (approved):

ekrembal (approved):

PR #874, opened 2025-06-24
Remove enable_hcp env var

atacann:
Remove ENABLE_HEADER_CHAIN_PROVER, instead use automation feature to determine if hcp should be run.

HCP is useless without automation anyway, and automation is useless in case of a challenge if hcp is not running.

Why? So verifiers that have and do not have automation can have the same config. (No-automation with hcp enabled crashes if firewall blocks bonsai requests)

Enables `verifier_new_check_header_chain_proof` test


core/src/verifier.rs
```diff
@@ -220,15 +221,11 @@ where
             sessions: HashMap::new(),
         };
 
-        // TODO: Removing index causes to remove the index from the tx_sender handle as well
```
ceyhunsen, line 223: I remember adding xonlypk to this handle in a PR which was merged if I am not mistaken. But there are no such change now. I suggest adding that change or leaving this todo.

ceyhunsen (changes requested):
rest looks good

core/src/verifier.rs
atacann: Sure, I added todo back because I kinda just want to delete it, that consumer id is not really a consumer id and is just used in a tracing debug log, later maybe

atacann: It might potentially be helpful in our own tests, but I never needed it. In deployment every verifier/operator has their own separate logs and verifiers have separate db's so we don't need an identifier.

mmtftr:
Thanks for the change, I was wondering why we had that ENABLE variable anyway.

mmtftr (commented):
I would also add in the description that this PR enables the `verifier_new_check_header_chain_proof` test

PR #892, opened 2025-06-30
Add data tests for watchtower challenges with various challenge txs

ozankaymak:
This PR introduces the production of different types of watchtower challenge transactions.
The reason of change for the Version field of the watchtower transaction is to be able to send large transactions without problem.


core/src/operator.rs
```diff
@@ -1528,6 +1528,81 @@ where
             }
         }
 
+        #[cfg(test)]
+        {
```
aoengin, line 1532: Can you please move this block to test.rs and make a call to keep production code clean?

core/src/builder/transaction/challenge.rs
```diff
@@ -92,21 +92,41 @@ pub fn create_watchtower_challenge_txhandler(
     #[cfg(test)]
```
aoengin, line 92: same, mission: impossible - keep prod code as clean as possible

circuits-lib/src/bridge_circuit/lc_proof.rs
```diff
@@ -30,19 +30,25 @@ mod tests {
     #[test]
     fn test_lc_proof_verifier() {
```
aoengin, line 31: we need to find a way to test the actual method

aoengin (changes requested):

bridge-circuit-host/src/bridge_circuit_host.rs
```diff
@@ -597,4 +631,74 @@ mod tests {
...
+    fn test_bridge_circuit_with_large_input() {
+        eprintln!("\x1b[31mPlease update test data if the elf files are changed. Run the tests on bridge_circuit_test_data.rs to update the test data.\x1b[0m");
+        let input_bytes: &[u8] =
+            include_bytes!("../bin-files/bch_params_challenge_tx_with_large_annex.bin");
+        let mut bridge_circuit_host_params: BridgeCircuitHostParams =
+            borsh::from_slice(input_bytes)
+                .expect("Failed to deserialize BridgeCircuitHostParams from file");
+
+        // Now add the removed witness element back to the watchtower inputs
+        for watchtower_input in &mut bridge_circuit_host_params.watchtower_inputs {
```
atacann, line 662: whats the reason for adding extra witness here? and where was it removed (according to comment)

ozankaymak: We want to check whether the proving fails since the operator should only put the signature, not the whole witness

atacann (approved):

aoengin (approved):

Issue #865, opened 2025-06-23
Disable Operator::Withdraw when Automation Disabled


mmtftr:
As the title says. Also, use TxSender instead of send_raw_transaction

PR #893, opened 2025-06-30
Disable withdrawal on no-automation - need reviews

atacann:
- Disables withdrawals if automation feature is not used, calling withdraw() grpc endpoint returns an error with a message saying automation is not enabled.
- Uses txsender RBF for payouts generated in withdraw(). Because of this change withdraw() endpoint doesn't return the payout txid anymore (because the txid can change depending on current fees inside TxSender, before we funded one tx, sent it to mempool and returned this txid). Because of this the tests that relied on the txid being returned were changed.
- Deletes withdrawalfinalized grpc endpoint and related proto struct, this endpont was unused

#865


circuits-lib/src/bridge_circuit/mod.rs
```diff
@@ -624,7 +624,7 @@ pub fn journal_hash(
 
 /// Retrieves the first output of a transaction that is an OP_RETURN script. Used to retrieve
 /// the operator's X-only public key from the OP_RETURN output in the payout transaction.
-fn get_first_op_return_output(tx: &CircuitTransaction) -> Option<&TxOut> {
+pub fn get_first_op_return_output(tx: &CircuitTransaction) -> Option<&TxOut> {
```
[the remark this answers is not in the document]
atacann: Yeah maybe it might but only when we recompile the elfs. But yes maybe we should also have a ci job that regenerates elfs and checks if method id changed or something like that

core/src/rpc/clementine.proto
```diff
@@ -270,7 +268,7 @@ message WithdrawErrorResponse {
 
 message WithdrawResult {
   oneof result {
```
mmtftr, line 270: why do we need this type? we could've reused gRPC Status as the error variant.

I think it should be an empty response from Withdraw, and we can use the Status variants to indicate errors. We have support for wrapping errors with `Status` variants now, if you wrap an error with `.wrap_err(Status::invalid_argument("..."))` and return it, the `From` trait will return the proper error when converting from `BridgeError` back to `Status`.

core/src/verifier.rs
```diff
@@ -1628,7 +1627,7 @@ where
             }
             let payout_tx_idx = payout_tx_idx.expect("Payout tx not found in block cache");
             let payout_tx = &block.txdata[*payout_tx_idx];
-            // Find the output that contains OP_RETURN
+            // Find the first output that contains OP_RETURN
```
mmtftr, line 1630: use get_first_op_return_output here too

mmtftr (changes requested):

mmtftr (approved):

PR #898, opened 2025-07-01
Implement Bridge-Specific Tests

aoengin:


core/src/operator.rs
```diff
@@ -1369,6 +1386,23 @@ where
...
+        #[cfg(test)]
+        let blockhashes_serialized = {
+            if self
+                .config
+                .test_params
+                .generate_varying_total_works_invalid_total_work
+            {
+                block_hashes
+                    .iter()
+                    .take((payout_block_height + 1) as usize)
```
atacann, line 1398: this wont work properly if genesis height is not 0, but its fine because its just used in generation test, just so you are aware

atacann (commented):
Looks fine from just a quick look, I suggest you merge the base branch with main then merge this with the base branch to get a better git diff

PR #900, opened 2025-07-01
Test prereq ci cache

atacann:
Use CI cache for bitvm_cache and bitcoin 29.
With this the test prerequisite consistently takes about 1min for all tests.

I also increased poll delays for tasks, I believe this didn't affect the test runtime negatively but might decrease flakiness. Most likely it does nothing though.


ceyhunsen (approved):
didnt realize you ran the ci again to test these changes, thats why i started ci again

PR #901, opened 2025-07-01
Add state root check for LCP

### LINKED ISSUE #897: Double check LCP state root with Citrea node state root

ceyhunsen:
# Description

Adds get_l2_block_by_number Citrea endpoint for checking state root against LCP's state root.

## Linked Issues

- Closes #897

## Testing

Current tests pass


ceyhunsen:
Will do the same PR to the audit branch, once this is approved and merged

core/src/test/common/citrea/client_mock.rs
```diff
@@ -189,21 +189,23 @@ impl CitreaClientT for MockCitreaClient {
...
-    ) -> Result<(u64, u64), BridgeError> {
+    ) -> Result<(u64, u64, [u8; 32]), BridgeError> {
         Ok((
             if block_height == 0 {
                 0
             } else {
                 block_height - 1
             },
             block_height,
+            [0; 32],
```
ozankaymak, line 208: Why do we return [0u8; 32] here?

core/src/verifier.rs
```diff
@@ -2398,13 +2398,21 @@ where
...
-            l2_height_end,
-            block_height,
-        )
-        .await?;
+
+        if self
+            .citrea_client
+            .check_state_root(l2_height_end, state_root)
+            .await
+            .is_ok()
```
atacann, line 2406: I think we should return with error, it can cause problems if state root doesnt match but we advance to new block

core/src/test/common/citrea/client_mock.rs
ceyhunsen: its the mock interface. so, value shouldnt really matter as long as it is matched with the lcp's state root

core/src/verifier.rs
ceyhunsen: https://github.com/chainwayxyz/clementine/pull/901/commits/85f364e66f32b3a13a6770544d33189325665d68 currently we have some problems with tests, i believe. because for some heights, citrea doesn't return a real value (only returns 0). so for those heights, this causes endless loop as seen in the commit's ci/cd run.

atacann: I tried additional_disprove_script_test_disrupted_latest_block_hash and citrea_deposit_and_withdraw_e2e_non_zero_genesis_height locally and there were no mismatches between lcp and l2 state root so its weird. The tests didn't fail

Issue #808, opened 2025-06-03
Better fee estimation using third party providers


ceyhunsen:
### Proposal Description

Currently, `estimate_smart_fee` Bitcoin RPC call is used to estimate fee rate. This is not very reliable and third party options should be considered, like mempool.space API.

ceyhunsen:
closed with #932

PR #904, opened 2025-07-02
Remove crate level allow unuseds

ceyhunsen:
# Description

There was a crate level `allow`s which was causing problems.

### DIFF
--- a/core/src/builder/transaction/challenge.rs
+++ b/core/src/builder/transaction/challenge.rs
@@ -305,7 +305,7 @@ pub fn create_operator_challenge_ack_txhandler(
 pub fn create_disprove_txhandler(
     kickoff_txhandler: &TxHandler,
     round_txhandler: &TxHandler,
-    paramset: &'static ProtocolParamset,
+    _paramset: &'static ProtocolParamset,
 ) -> Result<TxHandler, BridgeError> {
     Ok(TxHandlerBuilder::new(TransactionType::Disprove)
         .with_version(Version::TWO)
--- a/core/src/musig2.rs
+++ b/core/src/musig2.rs
@@ -12,13 +12,9 @@ use bitcoin::{
 use eyre::Context;
 use lazy_static::lazy_static;
 use secp256k1::{
-    ffi::{
-        MusigAggNonce, MusigKeyAggCache, MusigPartialSignature, MusigPubNonce, MusigSecNonce,
-        MusigSession,
-    },
     musig::{
-        new_nonce_pair, AggregatedNonce, AggregatedSignature, KeyAggCache, PartialSignature,
-        PublicNonce, SecretNonce, Session, SessionSecretRand,
+        new_nonce_pair, AggregatedNonce, KeyAggCache, PartialSignature, PublicNonce, SecretNonce,
+        Session, SessionSecretRand,
     },
     rand::Rng,
     Scalar, SECP256K1,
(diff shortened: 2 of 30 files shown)


core/src/musig2.rs
```diff
@@ -195,7 +191,7 @@ pub fn aggregate_partial_signatures(
 /// <https://medium.com/blockstream/musig-dn-schnorr-multisignatures-with-verifiably-deterministic-nonces-27424b5df9d6#e3b6>.
 pub fn nonce_pair(
     keypair: &Keypair,
-    mut rng: &mut impl Rng,
+    _rng: &mut impl Rng,
```
atacann, line 194: We can remove this

core/src/builder/transaction/challenge.rs
```diff
@@ -305,7 +305,7 @@ pub fn create_operator_challenge_ack_txhandler(
 pub fn create_disprove_txhandler(
     kickoff_txhandler: &TxHandler,
     round_txhandler: &TxHandler,
-    paramset: &'static ProtocolParamset,
+    _paramset: &'static ProtocolParamset,
```
atacann, line 308: can remove this too

core/src/states/task.rs
```diff
@@ -56,23 +56,13 @@ impl BlockHandler for QueueBlockHandler {
 /// A task that fetches new blocks from Bitcoin and adds them to the state management queue
 #[derive(Debug)]
 pub struct BlockFetcherTask<T: Owner + std::fmt::Debug + 'static> {
-    /// The database to fetch events from
```
atacann, line 59: So this struct is just not used? Lets delete it but we need to confirm

core/src/tx_sender/mod.rs
```diff
@@ -793,7 +788,7 @@ mod tests {
         dbtx.commit().await?;
 
         // Get the current fee rate and increase it for RBF
-        let current_fee_rate = tx_sender._get_fee_rate().await?;
+        // let current_fee_rate = tx_sender._get_fee_rate().await?;
```
atacann, line 791: Unrelated here but why is get_fee_rate unused? I checked it and FeeRate is set to 1 by default inside txsendertask... . Lets fix that

atacann: Same with bump_fees_of_fee_payer_txs. Should it be used somewhere? @ekrembal @mmtftr

atacann (changes requested):
Just small comments and txsender stuff not related to this pr

core/src/tx_sender/mod.rs
ceyhunsen: if i remember correctly, returned fee rate is a bit off. #808

core/src/builder/transaction/challenge.rs
ceyhunsen: https://github.com/chainwayxyz/clementine/pull/904/commits/e17f7a4430ba65f6705f43aebf56f040757d9d71

core/src/states/task.rs
atacann: yes but its not actually being used. Its only purpose is to call FinalizedBlockFetcherTask::new on bitcoin syncer

ceyhunsen: Tried to delete this and convert `impl` to `FinalizedBlockFetcherTask` but end up having lots of trait errors. A fix will be messy. So, better to have it like this IMO.

PR #906, opened 2025-07-03
Risc0 circuits overhaul

ozankaymak:
# Description

## Linked Issues

- Closes # (issue, if applicable)
- Related to # (issue)

## Testing

added unit tests?

## Docs

have the docs been updated?

### DIFF
--- a/circuits-lib/Cargo.toml
+++ b/circuits-lib/Cargo.toml
@@ -12,7 +12,6 @@ risc0-groth16 = { workspace = true }
 
 borsh = { workspace = true, features = ["derive"] }
 bitcoin = { workspace = true, features = ["rand-std", "serde"] }
-ripemd = { workspace = true }
 
 ark-bn254 = { workspace = true }
 ark-ff = { workspace = true }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1983,7 +1983,6 @@ dependencies = [
  "num-bigint 0.4.6",
  "num-traits",
  "once_cell",
- "ripemd",
  "risc0-groth16",
  "risc0-zkvm",
  "serde",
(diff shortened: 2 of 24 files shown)


ekrembal (dismissed):

core/src/test/bridge_circuit_test_data.rs
```diff
@@ -0,0 +1,582 @@
+use super::common::citrea::get_bridge_params;
```
mmtftr, line 1: add some module docs so that it's clear these are not tests but are just to generate the test_data and that they must be rerun when ... (idk when, probably whenever the circuit method id changes or there is some deposit breaking change)

circuits-lib/Cargo.toml
```diff
@@ -49,4 +48,5 @@ lazy_static = { workspace = true }
 citrea-sov-rollup-interface = { workspace = true }
 
 [features]
-not_patched = []
+default = []
+use-test-vk = []
```
mmtftr, line 52: would appreciate a short comment here, what's a test-vk (ie. what is the difference of the circuit that uses this vk)

core/src/operator.rs
```diff
@@ -1368,6 +1385,23 @@ where
...
+        #[cfg(test)]
+        let blockhashes_serialized = {
+            if self
+                .config
+                .test_params
+                .generate_varying_total_works_invalid_total_work
+            {
+                block_hashes
+                    .iter()
+                    .take((payout_block_height + 1) as usize)
```
atacann, line 1397: payout_block_height + 1 - paramset.genesis_height so that in the future if we change default test setup somehow you dont debug this

atacann (approved):

mmtftr (approved):
thanks

PR #912, opened 2025-07-08
Add a full-stack Docker compose and update readme

### LINKED ISSUE #608: README TODOs
### Proposal Description

Readme needs small updates:

- [ ] Way to use Dockerhub image to start a server
- [x] PostgreSQL and RiscZero installations not only required in testing

ceyhunsen:
# Description

Adds

- Downloading BitVM cache
- Starting a server using DockerHub image
- Starting a server using provided compose file

steps to readme.

## Linked Issues

- Closes #608


ceyhunsen:
at this point, both testnet and regtest docker compose files do work

scripts/docker/docker-compose.regtest.yml
mmtftr, line 1: I think we need to rename this to `docker-compose.full.regtest.yml` and the below testnet file to `docker-compose.verifier.testnet.yml` as it should be clear that these run different subsets of the bridge.

scripts/docker/configs/regtest/.env.regtest
```diff
@@ -0,0 +1,79 @@
...
+DBG_PACKAGE_HEX=1
+RISC0_SKIP_BUILD=0
+RISC0_DEV_MODE=0
+BRIDGE_CIRCUIT_METHOD_ID_CONSTANT=e246ef42e7795aa55cf0f6677cbabb78dc1fc461c20c2addd8ff70c8c8d019db
+CA_CERT_PATH=/certs/ca/ca.pem
+SERVER_CERT_PATH=/certs/server/server.pem
+SERVER_KEY_PATH=/certs/server/server.key
+CLIENT_CERT_PATH=/certs/server/server.pem
+CLIENT_KEY_PATH=/certs/server/server.key
+AGGREGATOR_CERT_PATH=/certs/server/server.pem
```
mmtftr, line 68: this is not the correct setup for certificates. All entities other than the aggregator should have:


```suggestion
SERVER_CERT_PATH=/certs/server/server.pem
SERVER_KEY_PATH=/certs/server/server.key
CLIENT_CERT_PATH=/certs/client/client.pem
CLIENT_KEY_PATH=/certs/client/client.key
AGGREGATOR_CERT_PATH=/certs/aggregator/aggregator.pem
```

and aggregator should override just its client certificate

```env
CLIENT_CERT_PATH=/certs/aggregator/aggregator.pem
CLIENT_KEY_PATH=/certs/aggregator/aggregator.key
```

mmtftr (changes requested):

scripts/docker/docker-compose.regtest.yml
ceyhunsen: sounds good https://github.com/chainwayxyz/clementine/pull/912/commits/2ab453aeb1d41c7702df34ce36c84a02b663fcd2

mmtftr:
Not sure if this is out of scope but I think we should be using the `rust` Docker image instead of installing rust from scratch in a generic `ubuntu` image. `rust:1.85.0` would be better. (we can then remove rust's prerequisite installation steps and save on space)

mmtftr:
Also the `.dockerignore` currently includes everything (including the `target` directory) which is not recommended. I tested and found that this dockerignore works correctly on my local machine, can we update it to this?

```gitignore
# exclude everything
*

# include source files
!/circuits-lib
!/.cargo
!/risc0-circuits
!/bridge-circuit-host
!/core
!/scripts
!Cargo.lock
!Cargo.toml

# include for vergen constants
!/.git

# include licenses
!LICENSE-*
!LICENSE

# exclude bitvm cache (for local builders who forgot to delete it)
**/bitvm_cache.bin
**/bitvm_cache_dev.bin
```

scripts/docker/Dockerfile
mmtftr, line 1: thanks, can we further use a multi-stage build to reduce our image size? We don't need rust build tools in the final container, I was using this:

```Dockerfile
# Set up OS
FROM rust:1.85.0

# Compile Clementine
WORKDIR /clementine
COPY . .
RUN cargo +1.85.0 build --release --bin clementine-core

FROM ubuntu:22.04

COPY --from=0 /clementine/target/release/clementine-core /clementine-core

# Set up Clementine
ENTRYPOINT [ "/clementine-core" ]
EXPOSE 17000
```

ceyhunsen:
These were out of scope but because they were small changes, I added them.

- https://github.com/chainwayxyz/clementine/pull/912/commits/4107a97317df3f1c9a70a4727b1426a0958750b5
- https://github.com/chainwayxyz/clementine/pull/912/commits/88e34adf4c73e0ed5c6b09e10ce04bfed7d49d86

Also, all the Rust images have reported vulns: https://hub.docker.com/layers/library/rust/1.85.0/images/sha256-9ee2e8fc6db6e563e829ad9b772d7e912f539c6ca3fcc1dae9457510496e072d. Not sure if it effects us.

scripts/docker/Dockerfile
ceyhunsen: https://github.com/chainwayxyz/clementine/pull/912/commits/67263b728798b3229a599b64e92f587ed98e0459

scripts/docker/configs/regtest/.env.regtest
ceyhunsen: i am trying to test this but having other issues blocking it rn.

i copied this config from `run.sh`. why do we do that there, than?

ceyhunsen: https://github.com/chainwayxyz/clementine/pull/912/commits/36c0f24d95d881152a782ed966e4eeb4dfc28405

mmtftr: well, i think we just wrote it like that since we were having other issues at the time.

PR #915, opened 2025-07-09
Add Corrupted Public Input Test and Randomise Corrupted Assert Test

### LINKED ISSUE #873: Remove the code duplication for e2e (bitvm included) tests
### Proposal Description

Investigate and find a way to remove the code duplication in e2e testing, namely in these files:
- core/src/test/additional_disprove_scripts.rs
- core/src/test/bitvm_disprove_scripts.rs
- core/src/test/watchtower_challenge.rs

aoengin:
- Resolved #873.
- Added a new test for the corrupted public input case.
- Randomized the corrupted assert test.

### DIFF
--- a/core/src/operator.rs
+++ b/core/src/operator.rs
@@ -1561,30 +1561,7 @@ where
 
         #[cfg(test)]
         {
-            use rand::Rng;
-            if self.config.test_params.corrupted_asserts {
-                let mut rng = rand::thread_rng();
-
-                if rng.gen_bool(0.5) {
-                    let i = rng.gen_range(0..asserts.1.len());
-                    let j = rng.gen_range(0..asserts.1[i].len());
-                    tracing::info!("Disrupting asserts commit 1 with i: {}, j: {}", i, j);
-
-                    asserts.1[i][j] ^= 0x01;
-                } else {
-                    let i = rng.gen_range(0..asserts.2.len());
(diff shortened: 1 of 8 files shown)


ozankaymak (dismissed):

core/src/test/bitvm_disprove_scripts.rs
```diff
@@ -507,228 +563,100 @@ impl DisproveTest {
...
-        };
+                let disprove_timeout_tx = rpc.client.get_raw_transaction(&txid, None).await?;
 
-        tracing::info!(
-            "Disprove outpoint: {:?}, txid: {:?}",
-            disprove_outpoint,
-            kickoff_txid
-        );
+                assert!(
+                    disprove_timeout_tx.input[1].previous_output == kickoff_finalizer_out,
```
mmtftr, line 599: is this possible?

core/src/operator.rs
```diff
@@ -1561,9 +1561,29 @@ where
 
         #[cfg(test)]
         {
+            use rand::Rng;
```
mmtftr, line 1564: I would prefer to have as little test code as possible in our production code. Can we move this logic into the testparams impls, and call with `self.config.test_params.maybe_corrupt_asserts(&mut asserts)`

mmtftr: same for below?

mmtftr (dismissed):

PR #921, opened 2025-07-16
Upload deposit state artifact

atacann:
If test_bridge_contract_change fails CI generates and uploads the state file, devs can download this and commit if needed instead of generating state locally.

Also fixed release unit and integration tests that were running on debug mode

After fixing release mode tests, noticed concurrent deposit tests were failing due to OOM. The reason was bitvm disprove scripts, so wrapped the place where bitvm disprove root hash is calculated in a mutex (also changed the fn to move instead of clone the scripts) which fixed it. The part in mutex takes about 1.5-2 seconds in gigapc so I don't think it will impact deposit performance


ozankaymak (commented):
LGTM

ekrembal (approved):

PR #930, opened 2025-07-22
Optimistic payout security module

ekrembal:


core/src/config/env.rs
```diff
@@ -106,6 +107,14 @@ impl BridgeConfig {
             None
         };
 
+        let opt_payout_verification_address = std::env::var("OPT_PAYOUT_VERIFICATION_ADDRESS")
```
ekrembal, line 110: Can we use Option<> config var instead of getting this from env here?

core/src/verifier.rs
```diff
@@ -1100,7 +1151,38 @@ where
...
+                let script_pubkey_bytes = output_script_pubkey.as_bytes().to_vec();
+                let params = ClementineOptimisticPayoutMessage {
+                    withdrawal_id: deposit_id,
+                    input_signature: input_sig_bytes.into(),
+                    input_outpoint_txid: outpoint_txid_bytes.into(),
+                    input_outpoint_vout: input_outpoint.vout,
+                    output_script_pubkey: script_pubkey_bytes.into(),
+                    output_amount: output_amount.to_sat(),
+                };
+                let eip712_hash = params.eip712_signing_hash(&DOMAIN);
```
ekrembal, line 1171: There is a code duplication here right?
We can use Verifier::recover_address_from_signature

PR #940, opened 2025-07-24
Wait for more blocks before sending latest blockhash tx

atacann:
Basically, after seeing the latest watchtower utxo being spent, operator waits for finality_depth more blocks before sending latest blockhash tx

Created a constant for citrea docker image, so that all tests can be upgraded at the same time by changing that const


ozankaymak:
I think the Operator should wait until the deadline - some constant # of blocks

ekrembal (approved):

PR #947, opened 2025-07-25
feat: add symmetric grpc limits on client/server

mmtftr:
# Description

Introduces symmetric configurable limitations on gRPC client/servers

### DIFF
--- a/core/src/config/env.rs
+++ b/core/src/config/env.rs
@@ -1,7 +1,7 @@
 //! # Environment Variable Support For [`BridgeConfig`]
 
 use super::BridgeConfig;
-use crate::{deposit::SecurityCouncil, errors::BridgeError};
+use crate::{config::GrpcLimits, deposit::SecurityCouncil, errors::BridgeError};
 use bitcoin::{address::NetworkUnchecked, secp256k1::SecretKey, Amount};
 use std::{path::PathBuf, str::FromStr};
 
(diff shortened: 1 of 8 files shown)


core/src/config/mod.rs
```diff
@@ -130,6 +130,27 @@ pub struct BridgeConfig {
...
+    /// gRPC client/server limits
+    #[serde(default = "default_grpc_limits")]
+    pub grpc: GrpcLimits,
+}
+
+#[derive(Debug, Clone, Deserialize, PartialEq)]
+pub struct GrpcLimits {
+    pub max_message_size: usize,
+    pub timeout_secs: u64,
+    pub tpc_keepalive_secs: u64,
```
ozankaymak, line 143: Typo: tpc -> tcp

core/src/servers.rs
```diff
@@ -176,6 +181,8 @@ where
         ServerAddr::Unix(ref socket_path) => {
             let server_builder = tonic::transport::Server::builder()
                 .layer(AddMethodMiddlewareLayer)
+                .timeout(Duration::from_secs(43200)) // 12 hours
```
ozankaymak, line 184: Is it possible to use these from config as well, instead of hardcoded constants? Or is it better this way?

core/src/config/env.rs
```diff
@@ -20,6 +24,28 @@ where
         .map_err(|e| BridgeError::EnvVarMalformed(env_var, format!("{:?}", e)))
 }
 
+impl GrpcLimits {
+    pub fn from_env() -> Result<Self, BridgeError> {
+        let defaults = default_grpc_limits();
+        Ok(GrpcLimits {
+            max_message_size: read_string_from_env_then_parse::<usize>("GRPC_MAX_MESSAGE_SIZE")
```
ceyhunsen, line 31: all of the env var equivalents of the bridgeconfig uses the same name but uppercase. but these has grpc_ suffix, which is not following the conventions.

either rename the bridgeconfig member name or the env var. because env vars should be instinctive.

ceyhunsen (changes requested):

PR #948, opened 2025-07-25
CMT-01 

mmtftr:
# Description

Implement new traits for easier extraction of xonly pks from btc addresses

Also uses the last20bytes traits when appropriate and possible

### DIFF
--- a/core/src/utils.rs
+++ b/core/src/utils.rs
@@ -2,8 +2,8 @@ use crate::builder::transaction::TransactionType;
 use crate::errors::BridgeError;
 use crate::operator::RoundIndex;
 use crate::rpc::clementine::VergenResponse;
-use bitcoin::{OutPoint, TapNodeHash, XOnlyPublicKey};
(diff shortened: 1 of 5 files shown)


core/src/utils.rs
```diff
@@ -390,14 +391,50 @@ pub struct RbfSigningInfo {
...
 pub trait Last20Bytes {
-    fn last_20_bytes(self) -> [u8; 20];
+    fn last_20_bytes(&self) -> [u8; 20];
+}
+
+pub trait TryLast20Bytes {
+    fn try_last_20_bytes(self) -> Result<[u8; 20], BridgeError>;
+}
+
+const fn larger_than_20<const N: usize>() {
```
atacann, line 401: unused fn I think

mmtftr: yeah it was from something else i tried

ceyhunsen (approved):

PR #966, opened 2025-07-28
CMT-23

aoengin:
- Add early input_signature verification in optimistic_payout() to prevent resource-intensive processing for invalid requests, mitigating potential DoS attacks.
- Additionally, add a P2TR check to the get_all_pubkeys function to prevent panics.


ozankaymak (approved):

core/src/rpc/aggregator.rs
```diff
@@ -876,17 +877,65 @@ impl ClementineAggregator for Aggregator {
...
             }
+
+            // Prepare input and output of the payout transaction.
+            let withdrawal_prevout = self
+                .rpc
+                .get_txout_from_outpoint(&input_outpoint)
+                .await
+                .map_to_status()?;
+
+            let user_xonly_pk =
```
atacann, line 888: Can you merge this with base branch then use the fn Efe wrote for CMT-1 to get xonly pk so that it doesn't panic if its not taproot

PR #969, opened 2025-07-28
CMT-28 Add evm_address to the op_return of challenge tx

ekrembal:
We use the secret key already in the config to reduce the secret management, this is really the edge case where the challenger sends the challenge tx automatically instead of manual sending

### DIFF
--- a/core/src/builder/transaction/challenge.rs
+++ b/core/src/builder/transaction/challenge.rs
@@ -3,7 +3,6 @@
 //! This module provides functions for constructing and challenge related transactions in the protocol.
 //! The transactions are: Challenge, ChallengeTimeout, OperatorChallengeNack, OperatorChallengeAck, Disprove.
 
-use crate::{builder, EVMAddress};
 use crate::builder::script::SpendPath;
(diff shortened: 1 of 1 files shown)


core/src/actor.rs
```diff
@@ -259,6 +262,16 @@ impl Actor {
         }
     }
 
+    pub fn get_evm_address(&self) -> Result<EVMAddress, BridgeError> {
```
ozankaymak, line 265: Is this the best way? I am not sure. Maybe just put the public key into the config as we do not necessarily have to have the secret key here

ekrembal: I've added my response in the pr comment

ekrembal: I've added my response in the pr comment

atacann (commented):
I feel like we should make evm addr non-optional, for example I think rn automated challenges do not have evm address in the tx (handle_kickoff -> create_and_sign_txs (here context is created with new_context_for_kickoff) -> create_txhandlers

ozankaymak (approved):

PR #1001, opened 2025-08-04
Cli reimbursement

atacann:
Creates new rpc endpoint + cli command for getting reimbursement manually. The rpc endpoint always prints the next tx to send (or error if it can't) to get a reimbursement given a deposit outpoint (maybe it should be movetx)

Enables withdraw endpoint on no automation operators again, uses fund_raw_transaction instead of txsender again as before


ceyhunsen (changes requested):

core/src/bin/cli.rs
```diff
@@ -520,16 +528,80 @@ async fn handle_operator_call(url: String, command: OperatorCommands) {
...
+            }
+
+            println!(
+                "Getting kickoff txs for outpoint {}:{}",
+                deposit_outpoint_txid, deposit_outpoint_vout
+            );
+            let response = operator
+                .get_reimbursement_txs(Request::new(Outpoint {
+                    txid: Some(clementine_core::rpc::clementine::Txid {
+                        txid: hex::decode(deposit_outpoint_txid).expect("Failed to decode txid"),
```
ekrembal, line 558: This needs to be reversed

ekrembal (approved):
lgtm,
tested with Docker
Tested that the operator can't propogate Round Tx more than once

PR #1002, opened 2025-08-04
Add CLI command for cpfp

ekrembal:


core/src/bin/cli.rs
```diff
@@ -765,6 +780,230 @@ async fn handle_aggregator_call(url: String, command: AggregatorCommands) {
...
+            // if !signed_tx.complete {
+            //     println!("Failed to sign child transaction");
+            //     return;
+            // }
+
+            let mut new_psbt = Psbt::from_str(&signed_tx.psbt).expect("Failed to parse PSBT");
+
+            new_psbt.inputs[0] = bitcoin::psbt::Input {
+                witness_utxo: Some(TxOut {
+                    value: Amount::from_sat(0),
```
atacann, line 953: Copy p2a_vout here as TxOut because value might not be 0 (for kickoff tx for example)

atacann:
SendTxWithCpfp is currently an aggregator command but it doesn't actually depend on any clementine nodes

PR #1013, opened 2025-08-06
Security mod for withdrawals

atacann:
- Same as opt payouts, add an sig verification check for withdraw() rpcs of operators.
- Add option to specify which operators to send withdraw() to for aggregator.withdraw()


ekrembal (dismissed):
lgtm

PR #1014, opened 2025-08-07
Rename ExtendedRpc to ExtendedBitcoinRpc

### LINKED ISSUE #786: Rename ExtendedRPC to BitcoinRPC
### Proposal Description

When the `ExtendedRPC` module was created, there was no other RPC related modules. Now that we have our own RPC endpoints, we should rename it. People confuse the two modules.

ozankaymak:
Closes #786 .


core/src/extended_rpc.rs
ceyhunsen, line 1: file name also needs to be changed

ceyhunsen (changes requested):

PR #1037, opened 2025-08-19
fix: cpfp reorg

### LINKED ISSUE #788: CPFP child transaction gets discarded if fee payer tx is also reorged
### Expected Behavior

`tx_sender` should re-send the package with the new child tx after a reorg

### Current Behavior

CPFP packages stuck in mempool because no fees

### Stps to Reproduce

1. Successfully send a CPFP tx on-chain
2. Reorg fee payer tx + CPFP package
3. Wait for fee payer tx to be on-chain again
4. Check that parent CPFP tx has no child and no fees

### Current Behavior

_No response_

### Detailed Description

CPFP packages should be saved in a temporary database until they are finalized. If a reorg is detected, it should wait until fee payer tx is on-chain and recreate CPFP package + send it.

atacann:
Closes #788
Fixes cpfp reorg issue that happened because a fee payer of a cpfp tx got reorged, that resulted in eviction of cpfp package in the mempool. After eviction it was never sent again by our node.

Solution I did is to just resend txs if they were not included in a block, after a new block arrives. This can fix some other issues too (some txs not getting sent by our rpc, but we never send it to rpc again if we send it successfully once)

- Adds migration support

### DIFF
--- a/core/src/extended_bitcoin_rpc.rs
+++ b/core/src/extended_bitcoin_rpc.rs
@@ -776,7 +776,7 @@ impl ExtendedBitcoinRpc {
     }
 
     /// A helper fn to safely mine blocks while waiting for all actors to be synced
-    /// If CitreaE2EData is provided and there are multiple DA nodes, it will additionally perform a reorg. Reorg will cause chain size to increase by number of reorgedd blocks in addition to block_num.
+    /// If CitreaE2EData is provided and there are multiple DA nodes, it will additionally perform a reorg. Reorg will cause chain size to increase by number of reorged blocks in addition to block_num.
     #[cfg(test)]
     pub async fn mine_blocks_while_synced<C: CitreaClientT>(
         &self,
--- a/core/src/test/common/tx_utils.rs
+++ b/core/src/test/common/tx_utils.rs
@@ -339,7 +339,7 @@ pub async fn create_tx_sender(
     use bitcoin::secp256k1::SecretKey;
 
     let sk = SecretKey::new(&mut rand::thread_rng());
-    let network = bitcoin::Network::Regtest;
+    let network = config.protocol_paramset().network;
     let actor: Actor = Actor::new(sk, network);
 
     let config = {


core/src/database/tx_sender.rs
```diff
@@ -583,6 +601,7 @@ impl Database {
         tx: Option<DatabaseTransaction<'_, '_>>,
         fee_rate: FeeRate,
         current_tip_height: u32,
+        latest_block_id: u32,
```
ceyhunsen, line 604: can you document this field

core/src/database/tx_sender.rs
```diff
@@ -647,7 +666,7 @@ impl Database {
                     -- Transaction must not be already confirmed
                     AND txs.seen_block_id IS NULL
                     -- Check if fee_rate is lower than the provided fee rate or null
-                    AND (txs.effective_fee_rate IS NULL OR txs.effective_fee_rate < $1);",
+                    AND (txs.effective_fee_rate IS NULL OR txs.effective_fee_rate < $1 OR txs.sent_block_id <> $3);",
```
ceyhunsen, line 669: comment is outdated

core/src/tx_sender/mod.rs
```diff
@@ -327,10 +327,11 @@ impl TxSender {
         &self,
         new_fee_rate: FeeRate,
         current_tip_height: u32,
+        latest_block_id: u32,
```
ceyhunsen, line 330: docs

core/src/tx_sender/task.rs
```diff
@@ -30,6 +30,7 @@ const POLL_DELAY: Duration = if cfg!(test) {
 pub struct TxSenderTask {
     db: Database,
     current_tip_height: u32,
+    latest_block_id: Option<u32>,
```
ceyhunsen, line 33: not sure here but can we make the dafault 0 instead of `None`? does it cause issues?

atacann: After a restart, while txsendertask didn't process any blocks yet, it will have highest_block_id 0 and try to send all txs and set their sent_block_id on db to 0 again. I don't think this breaks anything but that's why I did it. I don't think there is an issue leaving it an option

core/src/database/tx_sender.rs
```diff
@@ -633,12 +633,12 @@ impl Database {
     /// - Not in the cancelled list
     /// - Transaction itself is not already confirmed
     /// - Transaction and UTXO timelocks must be passed
-    /// - Fee rate is lower than the provided fee rate or null (deprecated)
+    /// - Fee rate is lower than the provided maximum fee rate (previous sends had a lower fee rate) or null (transaction wasn't sent before) OR the transaction was sent before, but the chain height increased since then, and the transaction is still not confirmed (accomplished by calling this fn with a dummy high fee rate)
```
mmtftr: with u32::MAX fee rate

mmtftr (approved):
minor doc comments, we can merge without

mmtftr (approved):

PR #1053, opened 2025-09-11
fix: anchor amount in cli cpfp

atacann:


core/src/bin/cli.rs
```diff
@@ -1007,7 +1007,7 @@ async fn handle_bitcoin_call(url: String, command: BitcoinCommands) {
 
             // Calculate package fee requirements
             let parent_weight = tx.weight();
-            let estimated_child_weight = bitcoin::Weight::from_wu(500);
```
mmtftr, line 1010: would appreciate comment on source of calculation (short summary like anchor input (xx WU) + fee payer (xx WU) -> change (xx WU) OR just a link to existing doc online)

mmtftr (approved):

PR #1057, opened 2025-09-12
fix: change amount in burn unused kickoff connectors

atacann:
- Fixes change amount in create_burn_unused_kickoff_connectors_txhandler (only relevant for standard bridges though)
- Adds underflow check in ready_to_reimburse_txhandler


core/src/builder/transaction/operator_collateral.rs
```diff
@@ -385,19 +388,24 @@ pub fn create_burn_unused_kickoff_connectors_txhandler(
...
             Sequence::from_height(1),
         );
     }
-    if !paramset.bridge_nonstandard {
+    if !paramset.bridge_nonstandard && input_amount >= paramset.anchor_amount() {
         // if we use standard tx's, kickoff utxo's will hold some sats so we can return the change to the change address
         // but if we use nonstandard tx's with 0 sat values then the change is 0 anyway, no need to add an output
         tx_handler_builder = tx_handler_builder.add_output(UnspentTxOut::from_partial(TxOut {
-            value: MIN_TAPROOT_AMOUNT,
+            value: input_amount - paramset.anchor_amount(),
```
ozankaymak, line 408: Is this definitely standard after subtraction?

PR #1073, opened 2025-09-15
fix: waiting for all watchtowers before sending assert

atacann:


core/src/tx_sender/client.rs
```diff
@@ -281,16 +283,8 @@ impl TxSenderClient {
                 // do not send watchtowet timeout if kickoff is already finalized
                 // which is done by adding kickoff finalizer utxo to cancel_outpoints
                 // this is not needed for any timeouts that spend the kickoff finalizer utxo like AssertTimeout
-                let kickoff_txid = related_txs
```
mmtftr, line 284: but why did we make this change

PR #1085, opened 2025-09-15
fix: block_cache_update stale data

atacann:
update_with_block kept stale data in cache, but we always used blockcache by creating a new one, so didn't catch this


mmtftr:
#1062 removes this function altogether and recreates state context from scratch. Maybe close this in favor of that?

PR #1114, opened 2025-09-17
fix(config): Check certain config and env variables if network is mainnet.

### LINKED ISSUE #1066: Add checks for test config vars that must be disabled in mainnet
To make misconfigurations impossible on mainnet:
- Stuff like client_verification=0 should be disabled in code (it panics on startup) if network is mainnet.
- If DISABLE_NOFN_CHECK=1 but network is mainnet, panic
- RISC0_DEV_MODE
- Think about other stuff

ceyhunsen:
# Description

Checks if certain configs or env vars are set if network is mainnet and returns error if so.

## Linked Issues

- Closes #1066

## Testing

Added a unit test that checks every scenario for a config.

### DIFF
--- a/core/src/config/mod.rs
+++ b/core/src/config/mod.rs
@@ -205,18 +205,17 @@ impl BridgeConfig {
 
         tracing::trace!("Using configuration file: {:?}", path);
 
-        let config = BridgeConfig::try_parse_from(contents)?;
-
-        config.check_mainnet_requirements()?;
-
-        Ok(config)
+        BridgeConfig::try_parse_from(contents)
     }
 
     /// Try to parse a `BridgeConfig` from given TOML formatted string and
     /// generate a `BridgeConfig`.
     pub fn try_parse_from(input: String) -> Result<Self, BridgeError> {
         match toml::from_str::<BridgeConfig>(&input) {
-            Ok(c) => Ok(c),
+            Ok(c) => {
+                c.check_mainnet_requirements()?;
+                Ok(c)
+            }
             Err(e) => Err(BridgeError::ConfigError(e.to_string())),
         }
     }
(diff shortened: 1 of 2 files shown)


core/src/config/mod.rs
```diff
@@ -212,10 +212,49 @@ impl BridgeConfig {
...
+    /// Checks various variables if they are correct for mainnet deployment.
+    fn check_mainnet_requirements(&self) -> Result<(), BridgeError> {
+        if self.protocol_paramset().network != Network::Bitcoin {
+            return Ok(());
+        }
+
+        let mut misconfigs = Vec::new();
+
+        if self.client_verification {
+            misconfigs.push("client_verification=true".to_string());
```
atacann, line 232: Here client_verification should be enabled for verifiers/operators, for aggregator we shouldn't force anything.

Also, for operators in mainnet we might want to force operator_collateral_funding_outpoint to be used in config @ekrembal . Also maybe for reimburse address too?

ekrembal: Yes, let's enforce OPERATOR_COLLATERAL_FUNDING_OUTPOINT for mainnet, i don't think we should enforce OPERATOR_REIMBURSEMENT_ADDRESS

core/src/config/mod.rs
```diff
@@ -216,6 +217,51 @@ impl BridgeConfig {
...
+    /// Checks various variables if they are correct for mainnet deployment.
+    pub fn check_mainnet_requirements(&self, actor_type: cli::Actors) -> Result<(), BridgeError> {
+        if self.protocol_paramset().network != Network::Bitcoin {
+            return Ok(());
+        }
+
+        let mut misconfigs = Vec::new();
+
+        if actor_type == cli::Actors::Operator {
+            if self.client_verification {
```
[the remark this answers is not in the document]
ceyhunsen: https://github.com/chainwayxyz/clementine/pull/1114/commits/13c87bad8a3c0b0fdadd2228a5333bfc9ef238b5

atacann:
LGTM but lets not merge for now in case we want to add other stuff too

atacann (approved):

PR #1134, opened 2025-09-22
Enhance Fee Rate Logic

aoengin:
- Exponential Backoff
- Retry Logic
- Min of Mempool Space and  Bitcoin RPC
- Hard Cap


core/src/tx_sender/mod.rs
```diff
@@ -121,96 +123,130 @@ impl TxSender {
...
+    /// *   **Mainnet and Testnet4:** Fetches fee rates from both Mempool Space API and Bitcoin Core RPC and takes the minimum.
+    /// *   **Hard Cap:** Applies a hard cap from configuration to prevent excessive fees.
+    /// # Fallbacks
+    /// *   If one source fails, it uses the other.
+    /// *   If both fail, it falls back to a default of 1 sat/vB.
     async fn get_fee_rate(&self) -> Result<FeeRate> {
-        match self.paramset.network {
+        match self.config.protocol_paramset.network {
             // Regtest and Signet use a fixed, low fee rate.
             Network::Regtest | Network::Signet => {
```
ekrembal, line 156: Let's remove this so that we can test our tx sender in signet better

PR #1139, opened 2025-09-23
Some circuit fixes

### LINKED ISSUE #1052: Add size check to mantissa
https://github.com/chainwayxyz/citrea/issues/2814
https://github.com/chainwayxyz/citrea/pull/2816

### LINKED ISSUE #1138: Use correct value for calculate work
Use `2**256` instead of U256::MAX

ozankaymak:
# Description
This PR:

- Adds mantissa size check
- Fixes the work calculation

## Linked Issues

- Closes #1052
- Closes #1138

## Docs
Updated the docs

### DIFF
--- a/circuits-lib/src/header_chain/mod.rs
+++ b/circuits-lib/src/header_chain/mod.rs
@@ -687,7 +687,8 @@ fn check_hash_valid(hash: &[u8; 32], target_bytes: &[u8; 32]) {
 /// Bitcoin measures cumulative proof-of-work as the sum of work done by all blocks.
 /// The work for a single block is inversely proportional to its target:
 /// work = 2 ** 256 / (target + 1)
-///
+/// The following implementation follows Rust-Bitcoin's approach:
+/// https://github.com/rust-bitcoin/rust-bitcoin/blob/eb17995e49b68831114cbc8bda14cbe72811c4b7/bitcoin/src/pow.rs#L533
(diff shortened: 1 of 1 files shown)


circuits-lib/src/header_chain/mod.rs
```diff
@@ -692,9 +704,25 @@ fn check_hash_valid(hash: &[u8; 32], target_bytes: &[u8; 32]) {
 ///
 /// * `U256` - The amount of work represented by this target
 fn calculate_work(target: &[u8; 32]) -> U256 {
```
atacann, line 706: is this btc core equivalent of this fn?

```
arith_uint256 GetBlockProof(const CBlockIndex& block)
{
    arith_uint256 bnTarget;
    bool fNegative;
    bool fOverflow;
    bnTarget.SetCompact(block.nBits, &fNegative, &fOverflow);
    if (fNegative || fOverflow || bnTarget == 0)
        return 0;
    // We need to compute 2**256 / (bnTarget+1), but we can't represent 2**256
    // as it's too large for an arith_uint256. However, as 2**256 is at least as large
    // as bnTarget+1, it is equal to ((2**256 - bnTarget - 1) / (bnTarget+1)) + 1,
    // or ~bnTarget / (bnTarget+1) + 1.
    return (~bnTarget / (bnTarget + 1)) + 1;
}
```

PR #1149, opened 2025-09-24
do not silently omit psbt outputs, return err

atacann:
Do not silently omit if Address::from_script doesn't work. I checked quickly if we can still call wallet_create_funded_psbt even if Address::from_script doesn't work but didn't find anything

### DIFF
--- a/core/src/tx_sender/rbf.rs
+++ b/core/src/tx_sender/rbf.rs
@@ -155,6 +155,7 @@ impl TxSender {
             estimate_mode: None,
         };
 
+        let mut omitted = 0usize;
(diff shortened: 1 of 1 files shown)


core/src/tx_sender/rbf.rs
```diff
@@ -176,13 +177,22 @@ impl TxSender {
                         out.value,
                     )),
                     Err(err) => {
-                        tracing::warn!("Failed to create address from script: {}", err);
+                        tracing::error!(
+                            "Failed to get address from scriptfor output of tx with txid {} for script: {}",
```
ozankaymak, line 181: Typo

ekrembal (approved):

PR #1157, opened 2025-09-29
Add db signature verification

atacann:
Adds signature verification for signatures retrieved from db with the txhandler we have constructed

### DIFF
--- a/core/src/actor.rs
+++ b/core/src/actor.rs
@@ -616,13 +616,33 @@ impl Actor {
                         sighash_type,
                     });
 
+                    let script_path_verify_sig = |sig: schnorr::Signature,
+                                                  xonly_public_key: XOnlyPublicKey|
+                     -> Result<(), BridgeError> {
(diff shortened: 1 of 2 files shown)


core/src/actor.rs
```diff
@@ -733,6 +779,28 @@ impl Actor {
...
+        sig: &schnorr::Signature,
+        sighash: TapSighash,
+        xonly_public_key: XOnlyPublicKey,
+        tweak_data: TapTweakData,
+        signature_id: &SignatureId,
+    ) -> Result<(), BridgeError> {
+        verify_schnorr(
+            sig,
+            &Message::from(sighash),
+            xonly_public_key,
```
mmtftr, line 794: add first few bytes in hex for xonly pk to the error details

core/src/actor.rs
mmtftr, line 599: add details on the kickoff data inside wrap_err wherever this is called

PR #1162, opened 2025-09-30
Fix bump fee calculate new fee rate logic

ozankaymak:


core/src/tx_sender/rbf.rs
```diff
@@ -45,27 +45,35 @@ impl TxSender {
         // Calculate original tx fee
         let original_tx_fee = self.get_tx_fee(&original_tx).await.map_err(|e| eyre!(e))?;
 
-        //println!("original_tx_fee: {}", original_tx_fee);
+        //tracing::debug!("original_tx_fee: {}", original_tx_fee);
 
         let original_tx_weight = original_tx.weight();
 
         // Conservative vsize calculation
         let original_tx_vsize = original_tx_weight.to_vbytes_floor();
```
atacann, line 53: This might have to be ceil(), or better maybe use sats per weight instead to not handle this

aoengin (approved):

PR #1171, opened 2025-10-03
Fix exponential backoff misuse

ozankaymak:


core/src/extended_bitcoin_rpc.rs
```diff
@@ -92,25 +92,34 @@ impl RetryConfig {
 
     pub fn get_strategy(&self) -> Box<dyn Iterator<Item = Duration> + Send> {
         // Clone the base strategy to get a fresh iterator with the same initial state
-        let strategy = (*self.base_strategy).clone();
+        let base_strategy = (*self.base_strategy).clone();
+        let base_delays = base_strategy.clone().collect::<Vec<_>>();
```
mmtftr: I think copilot has a point here, let's limit the base delays to 50 in case the strategy does not stop. We can use Iterator::take(n)

core/src/extended_bitcoin_rpc.rs
```diff
@@ -1072,6 +1096,7 @@ impl RpcApi for ExtendedBitcoinRpc {
         cmd: &str,
         args: &[serde_json::Value],
     ) -> std::result::Result<T, bitcoincore_rpc::Error> {
+        tracing::debug!("Calling Bitcoin RPC command: {}", cmd);
```
mmtftr: trace

core/src/extended_bitcoin_rpc.rs
```diff
@@ -66,17 +66,17 @@ pub struct RetryConfig {
...
+        initial_delay: u64,
         max_delay: Duration,
         max_attempts: usize,
         backoff_multiplier: u64,
         is_jitter: bool,
     ) -> Self {
         // Create the base strategy once
         let base_strategy = Arc::new(
-            ExponentialBackoff::from_millis(initial_delay.as_millis() as u64)
+            ExponentialBackoff::from_millis(backoff_multiplier)
```
mmtftr: this is also wrong in terms of the parameters of the tokio-retry struct.

Instead we likely have to calculate `factor` and `base` params manually. (note: `base` is the from_millis function's argument)

`base` will always correspond to the `backoff_multiplier` (because base multiplies with the delay on every iteration). `factor` is the multiplied factor to the internal delay before returning this iteration's delay.

To get `initial_delay == base_strategy.next()` (ie. first delay equal to the expected initial_delay) we'll have to calculate based on the implemented formula in the strategy struct.

From what I can see (please check), the `initial_delay` (ie. the first next() call's return value) will be `base * factor` milliseconds. Solving for base, we'd get `base = initial_delay / factor`. That's likely not a u64, we can either assert that initial_delay divides perfectly by factor or just rename the arguments to reflect this.

I think it'll also be good to add one test where we check that the initial_delay matches the expectation (in the case that factor divides initial_delay)

PR #1188, opened 2025-10-09
Check compatibility

### LINKED ISSUE #1173: Rpc call to get paramset
Add an rpc call to retrieve paramset of operators and verifiers

atacann:
- Adds rpc endpoints for retrieving variables that affect the signed bitcoin transactions + citrea chain ID
- Adds check on aggregator side on various rpcs like setup new_deposit that ensures all entities are compatible
- Moves hcp batch size out of protocolparams as it doesn't affect the transactions we sign

Here are whats being checked rn
```
pub struct CompatibilityParams {
    pub protocol_paramset: ProtocolParamset,
    pub security_council: SecurityCouncil,
    pub citrea_chain_id: u32,
    pub clementine_version: String,
    pub bridge_circuit_constant: [u8; 32],
    pub sha256_bitvm_cache: [u8; 32],
}
```
- For versions I check if major + minor version matches, we can change this behaviour, previously we didnt meaningfully change version number. For example we might force update minor or major version if bridge contract changes.
- Others except version are checked for exact equality (between entity and aggregators own config).

Closes #1173

### DIFF
--- a/core/build.rs
+++ b/core/build.rs
@@ -1,4 +1,8 @@
-use std::{env, path::Path, process::Command};
+use std::{
+    env,
+    path::{Path, PathBuf},
+    process::Command,
+};
 
 use vergen_git2::{BuildBuilder, CargoBuilder, Emitter, Git2Builder, RustcBuilder, SysinfoBuilder};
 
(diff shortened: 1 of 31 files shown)


core/src/compatibility.rs
```diff
@@ -0,0 +1,538 @@
...
+            "Failed to parse own Clementine version {}",
+            self.clementine_version
+        ))?;
+        let other_version = semver::Version::parse(&other.clementine_version).wrap_err(format!(
+            "Failed to parse other Clementine version {}",
+            other.clementine_version
+        ))?;
+        // allow different patch versions, but require major and minor versions to match for compatibility
+        // This ensures that only compatible protocol changes are accepted, while allowing bugfixes.
+        if own_version.major != other_version.major || own_version.minor != other_version.minor {
```
mmtftr: this is not correct for v0.x.x because in this case the second version number is actually the major version and the third is the minor. The parsing parses the major version as M.m.p without considering the special case of 0.M.m. 

For more reliable comparisons, we can instead use comparators which actually allow us to easily compare with own_version in a semver-compatible way.

```rs
Comparator::new(format!("^{own_version}")) // semver-compatible versions to own_version
```

I think it'll be good to write some tests for this in a separate function.

mmtftr (changes requested):

core/src/compatibility.rs
atacann: I think you mean its wrong for v1, 1.5 is not compatible with 1.6. I intended it to be this way but I guess we can use rust conventions, we have to increase major everytime bridge contract changes

PR #1218, opened 2025-11-11
Udocker-skopeo to remove docker-in-docker

atacann:
- Use skopeo pull - udocker run so that docker in docker is not needed
- Add spawn_blocking for proving work_only too
- When pulling image from dockerhub  (got through docker save or skopeo copy) if it has 2 identical layers, .tar files uses symlinks for all except one. And this breaks udocker load/create somehow (It didn't in my pc, but it did on debian docker image). That's why the fn that removes symlinks and copies layer files instead is needed.


bridge-circuit-host/src/docker.rs
```diff
@@ -221,7 +235,402 @@ const ID_BN254_FR_BITS: [&str; 254] = [
...
+                tar_file_path.to_string_lossy()
+            ))
+            .output()
+            .wrap_err("skopeo copy could not be executed")?;
+        if !pull_output.status.success() {
+            return Err(eyre!("skopeo copy failed: {:?}", pull_output));
+        }
+    }
+
+    let modified_tar_path = remove_symlinks_from_image_tar(&tar_file_path)?;
```
mmtftr: I think we should extract this to:

`async fn pull_or_load_image(image_digest: String) -> PathBuf` 

Then, for the Dockerfile.automation, we should embed the properly formatted tar files. Can we add a CLI command for cache_prover_images that'll run `pull_or_load_image` for all prover images, and cache them in the home directory.

bridge-circuit-host/src/docker.rs
```diff
@@ -221,7 +247,548 @@ const ID_BN254_FR_BITS: [&str; 254] = [
...
+    Ok(())
+}
+
+/// Pulls the image or loads the image from the cache if it exists. It ensures the config digest matches the given digest, or repulls the image if it doesn't. Returns the path to the modified tar file which deleted the symlinks.
+fn pull_or_load_image(
+    image_digest: &str,
+    container_name: &str,
+    image_config_digest: &str,
+) -> Result<PathBuf> {
+    let home_dir = std::env::home_dir().ok_or_else(|| eyre!("Failed to get HOME directory"))?;
```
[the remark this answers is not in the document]
atacann: its not deprecated

bridge-circuit-host/src/docker.rs
```diff
@@ -221,7 +262,546 @@ const ID_BN254_FR_BITS: [&str; 254] = [
...
+                );
+                fs::remove_file(&tar_file_path).wrap_err("Failed to delete local tar")?;
+            }
+        }
+    }
+
+    // 3. Pull with skopeo if needed
+    if need_pull {
+        tracing::info!("Pulling image via skopeo...");
+        let pull_output = Command::new("skopeo")
```
aoengin, line 567: We can check whether Skepeo is installed and provide a meaningful error message if it isn’t available.

bridge-circuit-host/src/docker.rs
```diff
@@ -221,7 +262,546 @@ const ID_BN254_FR_BITS: [&str; 254] = [
...
+/// - If digest verification fails, the cache is deleted to prevent use of a potentially compromised file.  
+///  
+/// # Error Conditions  
+/// - Returns an error if the tar file name or parent directory is invalid.  
+/// - Returns an error if file metadata cannot be read or if file operations (copy, delete) fail.  
+/// - Returns an error if digest verification fails or if the cache cannot be recreated.  
+///  
+/// # Returns  
+/// Returns the path to the modified (no-symlinks) tar file.  
+fn remove_symlinks_from_image_tar(path: &Path, expected_config_digest: &str) -> Result<PathBuf> {
```
aoengin, line 334: Since this function aims a specific bug, it's ok to keep it as is. But I would prefer replacing all symlinks if possible. 

Also see:

```
  // By closing the `TempDir` explicitly, we can check that it has
  // been deleted successfully. If we don't close it explicitly,
  // the directory will still be deleted when `tmp_dir` goes out
  // of scope, but we won't know whether deleting the directory
  // succeeded.
```

bridge-circuit-host/src/docker.rs
```diff
@@ -221,7 +262,546 @@ const ID_BN254_FR_BITS: [&str; 254] = [
...
+    let tmp_tar_path = tmp_dir.path().join("verified.tar");
+
+    // Copy from docker-archive to docker-archive - this forces skopeo to recompute all digests
+    // because it reconstructs the image structure. If the tar is tampered, this will fail
+    // or produce different digests.
+    tracing::debug!(
+        "Verifying tar file digest by copying to temporary docker-archive: {:?}",
+        tar_path
+    );
+    let copy_output = Command::new("skopeo")
```
aoengin, line 513: https://github.com/containers/skopeo/blob/main/docs/skopeo-copy.1.md

I could not find any explicit statements verifying the "Copy from docker-archive to docker-archive - this forces skopeo to recompute all digests" claim.

bridge-circuit-host/src/docker.rs
```diff
@@ -221,7 +262,458 @@ const ID_BN254_FR_BITS: [&str; 254] = [
...
+    }
+    Ok(())
+}
+
+/// Runs the prover container.
+/// skopeo is used to pull images instead of udocker pull because udocker pull doesn't support pulling from a sha digest.
+/// udocker is used instead of docker itself because it requires docker-in-docker to be set up if entities are run with docker.
+/// udocker commands are ran with --allow-root to avoid issues if the container is run as root. If it isn't run as root, --allow-root does nothing.
+fn run_prover_container(
+    image_digest: &str,
```
aoengin, line 628: Is it ok not to have mutex lock here?

aoengin (changes requested):

bridge-circuit-host/src/docker.rs
atacann: Its already acquired when this fn is called

bridge-circuit-host/src/docker.rs
atacann: it does

atacann: copy.go -> copy.Image -> copySingleImage -> copyLayers / copyConfig -> copyLayerFromStream / copyBlobFromStream -> newDigestingReader -> digestingReader.Read -> if actualDigest != d.expectedDigest

bridge-circuit-host/src/docker.rs
```diff
@@ -553,3 +1058,563 @@ pub fn to_decimal(s: &str) -> Option<String> {
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use tar::{Builder, EntryType, Header};
+
+    /// Test that pull_or_load_image succeeds for the STARK_TO_BITVM2 image.
+    /// This validates that STARK_TO_BITVM2_IMAGE_CONFIG_DIGEST is correct.
+    #[test]
+    fn test_pull_or_load_image_mainnet_bitvm2() {
```
aoengin, line 1070: Since this test has an external dependency, we can skip it for PRs not targeting main to reduce flakiness.

aoengin (changes requested):

.github/actions/test-prerequisites/action.yml
```diff
@@ -58,13 +58,32 @@ runs:
...
+        echo "/opt/udocker-1.3.17/udocker" >> $GITHUB_PATH
+
+    - name: Initialize udocker
+      shell: bash
+      run: |
+        sudo /opt/udocker-1.3.17/udocker/udocker --allow-root install
+
     - name: Install risc0
       shell: bash
-      env:
```
aoengin, line 63: why did we remove this?

atacann: Will add it back

bridge-circuit-host/src/docker.rs
atacann: Its not flaky though. You mean to reduce network usage due to pulling images?

aoengin (approved):

ekrembal (approved):

PR #1226, opened 2025-11-29
Add Circuit Check Workflow for Method ID Verification

aoengin:


.github/workflows/circuit_check.yml
```diff
@@ -0,0 +1,239 @@
...
+        id: cache-toolchain
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            ~/.risc0
+          key: ${{ runner.os }}-risc0-toolchain-${{ env.RISC0_TOOLCHAIN_VERSION }}
```
mmtftr: this should also depend on our rust version right?

.github/workflows/circuit_check.yml
```diff
@@ -0,0 +1,239 @@
...
+      - name: Verify bridge circuit method ID unchanged
+        run: |
+          if [ "${{ steps.pre_id.outputs.pre_id }}" != "${{ steps.post_id.outputs.post_id }}" ]; then
+            echo "Bridge method ID has changed from ${{ steps.pre_id.outputs.pre_id }} to ${{ steps.post_id.outputs.post_id }}"
+            exit 1
+          else
+            echo "Bridge method ID is unchanged: ${{ steps.pre_id.outputs.pre_id }}"
+          fi
+
+  header_chain_circuit_check:
```
mmtftr: can we use a matrix to write a single job instead of two?

```
jobs:
  circuit_check:
    strategy:
      matrix:
        circuit_name: ["header-chain", "work-only"]
  ...
 ```

File: .github/workflows/circuit_check.yml [hunk 3]
```diff
@@ -0,0 +1,239 @@
...
+
+env:
+  REPR_GUEST_BUILD: 1
+  BITCOIN_NETWORK: mainnet
+  RISC0_TOOLCHAIN_VERSION: v1
+
+jobs:
+  toolchain:
+    runs-on: ubicloud-standard-8
+    name: Warm RISC0 toolchain cache
```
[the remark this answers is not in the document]
aoengin: We won’t be changing the version too frequently. Since there are multiple jobs using the same tools, I thought it would be useful to cache them. I’m aware it doesn’t provide a big gain, though.

atacann (commented):
I think the workflow is fine but one issue is that we can't update rust version of clementine without this job giving errors. I t might be better to not use test-prerequisites, install a fixed version (1.88) here, and use that version in cargo build.

PR #1231, opened 2025-12-11
rpc for transferring funds from operator address to btc wallet

atacann:
- Adds rpc call to transfer utxos from operator address to btc  wallet connected to operator, checks first if these outpoints are collateral


ekrembal (dismissed):
LGTM, the last copilot comment makes sense

ekrembal (approved):

PR #1238, opened 2025-12-16
Tx sender bumping fixes

atacann:
- Do not assume funding inputs for rbf fee bumping are segwit, fill in inputs prevout info while bumping too (this was most likely the real problem, not segwit)
- Take into account min incremental fee while fee bumping cpfp
- Change effective fee in tx sender db to sat/kwu for more precision
- Bump fees if cpfp is not included in chain after some blocks
- Add option to disable version check for compatibility


core/src/tx_sender/mod.rs
```diff
@@ -42,6 +42,9 @@ mod task;
 pub use client::TxSenderClient;
 pub use task::TxSenderTask;
 
+/// Number of blocks after which a stuck transaction should be fee-bumped
+const FEE_BUMP_AFTER_BLOCKS: u32 = 10;
```
ekrembal: I think we can make this configurable

ekrembal (changes requested):
LGTM, i think we should make FEE_BUMP_AFTER_BLOCKS configurable

crates/clementine-tx-sender/src/rbf.rs
```diff
@@ -687,6 +715,17 @@ where
...
+            // Save the effective fee rate to the db
+            self.db
+                .update_effective_fee_rate(
+                    Some(&mut dbtx),
+                    try_to_send_id,
+                    effective_feerate,
+                    current_tip_height,
+                )
+                .await
+                .wrap_err("Failed to update effective fee rate")?;
```
aoengin: We can move db update below the if/else clause, by assigning an effective_feerate in both branches.

aoengin (changes requested):
LGTM

core/src/database/tx_sender.rs
```diff
@@ -1171,6 +1224,10 @@ impl clementine_tx_sender::TxSenderDatabase for Database {
         self.get_tx_debug_fee_payer_utxos(tx, tx_id).await
     }
 
+    async fn debug_inactive_txs(&self, fee_rate: FeeRate, current_tip_height: u32) {
+        self.debug_inactive_txs(fee_rate, current_tip_height).await
```
aoengin, line 1228: Inherent method is under test flag so prod code will recursively call the same trait method.

aoengin (changes requested):

core/src/test/common/mod.rs
```diff
@@ -388,13 +390,20 @@ pub async fn run_single_deposit<C: CitreaClientT>(
...
 
     let setup_start = std::time::Instant::now();
-    let mut aggregator = actors.get_aggregator();
-    let verifiers_public_keys: Vec<PublicKey> = aggregator
-        .setup(Request::new(Empty {}))
-        .await
-        .wrap_err("Failed to setup aggregator")?
-        .into_inner()
-        .try_into()?;
+    let strategy = ExponentialBackoff::from_millis(2).factor(500).take(3);
```
aoengin, line 393: Internal implementation of ExponentialBackoff is a bit confusing. Use RetryConfig instead.

core/src/compatibility.rs
```diff
@@ -64,24 +64,35 @@ impl CompatibilityParams {
...
-            self.clementine_version
-        ))?;
-        let other_version = semver::Version::parse(&other.clementine_version).wrap_err(format!(
-            "Failed to parse other Clementine version {}",
-            other.clementine_version
-        ))?;
-        let min_version = std::cmp::min(&own_version, &other_version);
-        let max_version = std::cmp::max(&own_version, &other_version);
-        let version_req =
+        if std::env::var("DISABLE_VERSION_CHECK")
```
aoengin, line 67: Should we add Mainnet check here not to silently ignore on Mainnet runs?

crates/clementine-tx-sender/src/lib.rs
```diff
@@ -661,6 +726,89 @@ where
...
+
+        // Get minimum fee increment rate from node for BIP125 compliance. Returned value is in BTC/kvB
+        let incremental_fee_rate = self
+            .rpc
+            .get_network_info()
+            .await
+            .map_err(|e| eyre::eyre!(e))?
+            .incremental_fee;
+        let incremental_fee_rate_sat_per_kvb = incremental_fee_rate.to_sat();
+        let incremental_fee_rate = FeeRate::from_sat_per_kwu(incremental_fee_rate_sat_per_kvb / 4);
```
aoengin, line 779: floor division, we should use ceil-div

aoengin (changes requested):

core/src/compatibility.rs
atacann: I think no, if versions are actually incompatible it will still fail on deposit stage anyway

ekrembal (approved):

aoengin (approved):

PR #1241, opened 2025-12-19
Use only op return for wt challenge

atacann:
- Update bitcoin ver to 30 in CI/readme (OP_RETURN > 80 bytes became standard)
- Update Citrea E2E version to latest (so that it uses bitcoin v30 + newer citrea)
- Use 1 144 byte OP_RETURN + anchor output and use cpfp to send watchtower challenges


ekrembal (dismissed):
Great work!
Should we remove all the things with Rbf resigning in tx sender and RbfSigningInfo?

ekrembal (dismissed):

bridge-circuit-host/src/utils.rs
```diff
@@ -296,20 +295,42 @@ pub fn calculate_succinct_output_prefix(method_id: &[u8]) -> [u8; 32] {
...
 }
 
-pub fn total_work_from_wt_tx(wt_tx: &Transaction) -> [u8; 16] {
-    let output = wt_tx.output[2].clone();
-    let mut instructions = output.script_pubkey.instructions();
-    if let Some(Ok(Instruction::Op(opcodes::all::OP_RETURN))) = instructions.next() {
-        if let Some(Ok(Instruction::PushBytes(data))) = instructions.next() {
-            let data_bytes = data.as_bytes();
-            let total_work: [u8; 16] = data_bytes[64..]
+pub fn total_work_from_wt_tx_test_util(wt_tx: &bitcoin::Transaction) -> [u8; 16] {
```
aoengin, line 298: #[cfg(test)]

core/src/builder/transaction/challenge.rs
```diff
@@ -9,10 +9,10 @@ use crate::builder::transaction::output::UnspentTxOut;
 use crate::builder::transaction::txhandler::{TxHandler, DEFAULT_SEQUENCE};
 use crate::builder::transaction::*;
 use crate::config::protocol::ProtocolParamset;
-use crate::constants::{MIN_TAPROOT_AMOUNT, NON_STANDARD_V3};
+use crate::constants::NON_STANDARD_V3;
```
aoengin, line 12: This is not nonstd anymore

bridge-circuit-host/src/bridge_circuit_host.rs
```diff
@@ -565,7 +565,8 @@ mod tests {
                 watchtower_input.watchtower_challenge_tx.output[2]
```
aoengin, line 746: This will start failing after updating the test data.

core/src/config/mod.rs
```diff
@@ -397,7 +398,7 @@ impl Default for BridgeConfig {
             )
             .expect("known valid input"),
 
-            operator_withdrawal_fee_sats: Some(Amount::from_sat(100000)),
+            operator_withdrawal_fee_sats: Some(NON_EPHEMERAL_ANCHOR_AMOUNT),
```
aoengin: Please add a comment explaining why we set it to NON_EPHEMERAL_ANCHOR_AMOUNT.

core/src/builder/transaction/challenge.rs
aoengin: outdated

aoengin (changes requested):

bridge-circuit-host/src/bridge_circuit_host.rs
aoengin: There are two more in the same file in addition to that one.

core/src/operator.rs
aoengin: This is a duplicate and is not related to this PR.

core/src/test/common/setup_utils.rs
```diff
@@ -245,6 +248,10 @@ pub async fn create_test_config_with_thread_name() -> BridgeConfig {
         citrea_rpc_url: handle.to_string(),
         ..Default::default()
     };
+
+    // 50% chance to enable annex tests. (Large annex is opt-in elsewhere.)
+    config.test_params.use_small_annex = rand::random::<bool>();
```
aoengin: This was mainly introduced to generate test inputs for the bridge circuit. It should be okay to keep, but it will write to the file every time it goes through send_asserts.

bridge-circuit-host/src/utils.rs
atacann: to  do that we need to remove the usage of the fn inside clementine core

core/src/builder/transaction/challenge.rs
atacann: We can rename in another PR if we want

aoengin (approved):

ekrembal (approved):

PR #1246, opened 2026-01-08
Fix TxSender flakiness with additional update queries

ekrembal:


core/src/database/tx_sender.rs
```diff
@@ -13,282 +13,168 @@ use clementine_tx_sender::{ActivatedWithOutpoint, ActivatedWithTxid};
...
+            WITH
+            u1 AS (
+                UPDATE tx_sender_activate_try_to_send_txids AS tap
+                SET seen_block_id = bs.id
+                FROM bitcoin_syncer_txs bst
+                JOIN bitcoin_syncer bs ON bst.block_id = bs.id
+                WHERE tap.txid = bst.txid
+                  AND tap.seen_block_id IS NULL
+                  AND bs.is_canonical = TRUE
+                RETURNING 1
```
aoengin, line 40: why did we add RETURNING 1, theye are not actively used?

core/src/database/tx_sender.rs
```diff
@@ -13,282 +13,168 @@ use clementine_tx_sender::{ActivatedWithOutpoint, ActivatedWithTxid};
...
-        // RBF transactions.
-        let common_ctes = r#"
-            WITH relevant_txs AS (
-                SELECT txid
-                FROM bitcoin_syncer_txs
-                WHERE block_id = $1
+        // Do all confirmation/unconfirmation updates in one round-trip.
+        // Postgres writable CTEs are executed in-order, so this preserves the
+        // original semantics while avoiding 14 separate UPDATEs.
+        let query = sqlx::query(
```
aoengin, line 29: seen_block_id can be temporarily set to NULL even when the tx is already in a new canonical block, because reconfirm only runs when seen_block_id IS NULL while unconfirm clears non-canonical references in the same sync cycle, reconfirm then happens on the next cycle.

aoengin (changes requested):

crates/clementine-tx-sender/src/cpfp.rs
```diff
@@ -446,11 +544,27 @@ where
...
+            if env::var("DBG_PACKAGE_HEX").is_ok() {
+                package
+                    .iter()
+                    .map(|tx| hex::encode(bitcoin::consensus::serialize(tx)))
+                    .collect::<Vec<_>>()
+            } else {
+                vec!["use DBG_PACKAGE_HEX=1 to print the package as hex".into()]
+            }
+        );
+
```
aoengin, line 567: duplicate, see lines 576-587

aoengin (changes requested):

core/src/database/tx_sender.rs
atacann: Thats fine I would say. Also, this code is getting deleted anyway

aoengin (dismissed):

atacann (dismissed):

PR #1249, opened 2026-01-09
Log every challenge tx, do not send challenge tx to txsender on mainnet

atacann:
- Closes #1243
- Do not send challenge txs automatically on mainnet or testnet4.
- Only add Kickoff timeout txs to txsender if the kickoff is challenged.


aoengin (approved):
LGTM

aoengin (changes requested):
core/src/rpc/clementine.proto (line 645) 
Checks if the kickoff tx is malicious and if so, try to send all necessary txs to punish the operator

Does not match the current behaviour.

ekrembal (approved):

aoengin (approved):

PR #1251, opened 2026-01-13
Tx sender decouple from btc syncer

### LINKED ISSUE #1250: RBF Psbt bump fee output changing S+AP signature
Psbt bump fee output is being reordered/subtracted etc, making S+AP signatures unusable. But S+AP RBF is only used in challenge tx which will be sent manually so not an issue for now, just informational.

atacann:
- Decouple txsender from clementine_core
- Add option for standalone running txsender
- Support citrea txs, add json rpc for citrea
- Closes #1250 also, for rbf's it conserves output ordering

- Didnt implement commit tx utxo fee estimation and giving the estimated reveal tx fee to utxo created by commit tx for now, I am worried about txsender somehow creating multiple commit txs if commit tx is detected as evicted or there is a problem with db that makes it hard to insert commit txid to the tables etc, if so many commit txs would be created. Disadvantage is that all reveal txs require one funding utxo at least. We can add the estimation later.


core/src/config/mod.rs
```diff
@@ -332,12 +332,49 @@ impl BridgeConfig {
...
+            TxSenderBitcoinRpcConfig, TxSenderConfig, TxSenderPostgresConfig,
+        };
+
+        TxSenderConfig {
+            network: self.protocol_paramset.network,
+            secret_key: self.secret_key,
+            private_da_key: None,
+            postgres: TxSenderPostgresConfig {
+                host: self.db_host.clone(),
+                port: u16::try_from(self.db_port).unwrap_or(5432),
```
aoengin: shouldn't this fail?

core/src/database/migrations/0002_txsender_seen_at_height_no_bitcoin_syncer_fk.down.sql
```diff
@@ -0,0 +1,122 @@
...
+DROP COLUMN IF EXISTS seen_at_height;
+
+ALTER TABLE tx_sender_cancel_try_to_send_txids
+DROP COLUMN IF EXISTS seen_at_height;
+
+ALTER TABLE tx_sender_activate_try_to_send_txids
+DROP COLUMN IF EXISTS seen_at_height;
+
+ALTER TABLE tx_sender_activate_try_to_send_outpoints
+DROP COLUMN IF EXISTS seen_at_height;
```
aoengin, line 41: drops seen_at_height without backfilling seen_block_id, so rollback loses txsender seen state.

core/src/database/migrations/0002_txsender_seen_at_height_no_bitcoin_syncer_fk.down.sql
```diff
@@ -0,0 +1,122 @@
+-- Rollback: restore legacy tx-sender `seen_block_id` columns and triggers.
```
aoengin, line 1: bitcoin_syncer(id) FK is not added back

core/src/database/migrations/0003_add_fee_paying_type_rbf_wtxid_grind.up.sql
```diff
@@ -0,0 +1,33 @@
...
+ADD COLUMN IF NOT EXISTS is_finalized boolean NOT NULL DEFAULT false;
+ALTER TABLE IF EXISTS tx_sender_activate_try_to_send_txids
+ADD COLUMN IF NOT EXISTS is_finalized boolean NOT NULL DEFAULT false;
+ALTER TABLE IF EXISTS tx_sender_cancel_try_to_send_outpoints
+ADD COLUMN IF NOT EXISTS is_finalized boolean NOT NULL DEFAULT false;
+ALTER TABLE IF EXISTS tx_sender_activate_try_to_send_outpoints
+ADD COLUMN IF NOT EXISTS is_finalized boolean NOT NULL DEFAULT false;
+-- Add unique constraint to txid column in tx_sender_try_to_send_txs table
+ALTER TABLE tx_sender_try_to_send_txs
+ADD CONSTRAINT uq_tx_sender_txid UNIQUE (txid);
```
aoengin, line 26: IF EXISTS is not natively supported for this clause as far as I know, but workarounds exist. We can add a DO block if you want.

core/src/rpc/clementine.proto
```diff
@@ -693,8 +694,16 @@ message SignedTxsWithType {
 }
 
 message RbfSigningInfoRpc {
```
aoengin, line 694:
```suggestion
message RbfSigningInfoRpc {
  uint32 vout = 1;
  uint32 tap_sighash_type = 2;

  oneof spend_path {
    RbfKeyPathSpend key_path = 3;
    RbfScriptPathSpend script_path = 4;
  }
}
```

Isn't this one clearer?

core/src/rpc/parser/mod.rs
```diff
@@ -98,17 +112,35 @@ impl TryFrom<RbfSigningInfoRpc> for RbfSigningInfo {
...
-            } else {
-                Some(
-                    TapNodeHash::from_slice(&value.merkle_root).wrap_err(eyre::eyre!(
-                        "Failed to convert merkle root bytes from rpc to TapNodeHash"
-                    ))?,
-                )
-            },
             vout: value.vout,
+            spend_path,
+            tap_sighash_type: TapSighashType::from_consensus_u8(value.tap_sighash_type as u8)
```
aoengin: u8::try_from before from_consensus_u8. 256 currently becomes 0 and can be accepted as a valid value.

crates/clementine-extended-rpc/src/client.rs
```diff
@@ -566,49 +568,48 @@ impl ExtendedBitcoinRpc {
...
         // Call Bitcoin Core's bumpfee RPC
         let bump_fee_result = match self
             .bump_fee(
                 &txid,
                 Some(&bitcoincore_rpc::json::BumpFeeOptions {
-                    fee_rate: Some(bitcoincore_rpc::json::FeeRate::per_vbyte(Amount::from_sat(
-                        new_fee_rate.to_sat_per_vb_ceil(),
-                    ))),
+                    fee_rate: Some(bitcoincore_rpc::json::FeeRate::per_kvbyte(
+                        Amount::from_sat(new_fee_rate.to_sat_per_vb_ceil()),
```
aoengin: should be to_sat_per_kvb

crates/clementine-extended-rpc/src/client.rs
aoengin: outdated and there are some other other outdated comments in this file please check

aoengin (changes requested):

core/src/database/migrations/0003_add_fee_paying_type_rbf_wtxid_grind.up.sql
atacann: It is already deployed on devnet, it works and changing would break the migrations

core/src/database/migrations/0002_txsender_seen_at_height_no_bitcoin_syncer_fk.down.sql
atacann: I will keep it like this, backfilling cna also cause problems. We do not except to rollback this anyway, if we rollback best solution would be to reset txsender sync. Best sol would be to not delete the seen_block_id column on the up migration but it's already deleted in devnet deployment

aoengin (approved):

ekrembal (approved):

PR #1259, opened 2026-02-05
Dump db of failed tests

atacann:
- Dump db of failed tests
- Change get_current_status rpc call log to debug
- For non standalone tests, uploading failed test logs didnt work properly, so I upload just the output of cargo test. (Its hard for tracing to log to different files in the same process, I couldn't find a way to differentiate which tokio threads belong to which test). This makes github web UI worse because stdout will include a lot more logs (warn->info log level change), but it will be easier while debugging.


aoengin (dismissed):

scripts/dump_failed_test_dbs.sh
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env bash
+set -euo pipefail
```
aoengin, line 2: set -e makes the script exit early when grep finds no match, because grep returns exit code 1

PR #1260, opened 2026-02-11
Add Slipstream Support

aoengin:
TODO: Handle CPFP "already submitted" error on slipstream before merge.


atacann (commented):
To reduce code duplication something like this can be done: You can create a wrapper for send_tx that takes a vec of transactions, if it has at least 2 it will send package, it it is 1 it will be send_raw_tx. Also if it has any nonstandard txs it will be sent with slipstream insteadd of rpc inside the wrapper. 

Getting modified feerate can also be done in try_to_send_unconfirmed_txs so that you dont do it twice in rbf and cpfp. You are currently updating effective feerate to base_fee_rate, not slipstream rate also right now. (RBF doesnt update effective feerate in main, it is added back in the open PRs).
If feerate was modified, you can check if ith current effective feerate, if it is <= current you can skip / or if is_tip_height_increased = true, do not check and just try to send it again (this was done to ensure if something was wrong with btc node/slipstream etc it will still try to send the same tx again to ensure its still in mempool). 

Also did you check if rbf/cpfp bumpfee can be done with slipstream (does it accept the bumped txs properly), or what if you send the exact same tx you sent before (because of is_tip_height_increased, txsender sends the same txs if it wasnt mined in the new block again). Slipstream has an api to check the status of a previously sent tx I believe, it might make sense to use that or just send the same tx again.

nonstandard module can be deleted I think, its just one fn

crates/clementine-tx-sender/src/maraslipstream_integration.rs
```diff
@@ -0,0 +1,253 @@
...
+        hex::encode(serialize(tx))
+    }
+
+    /// Attempts to submit a transaction to Slipstream when enabled for this tx.
+    ///
+    /// Returns:
+    /// - `Ok(Some(expected_txid))` when submitted (and txid check passed)
+    /// - `Ok(None)` when Slipstream is not applicable
+    /// - `Err(_)` when submission fails or txid mismatches
+    pub(crate) async fn submit_tx_via_slipstream(
```
atacann: Fn name can be changed because it doesn't definitely submit to slipstream. 
It would be good if this fn didn't log an error if the error just says this tx is already submitted or whatever it returns. Because txsender will re-attempt to send txs after every new block. Same with cpfp packages

PR #1262, opened 2026-02-20
Replacement cycling test

atacann:
Changes other than the new test:
- Remove standalone feature guard in txsender db run_migrations
- Fix cpfp change address so that a new address doesnt get created every create_child_tx call.
- Try to send all cpfp tx's every txsender loop, even if sent feerate is >= the current one.


crates/clementine-tx-sender/src/test_utils.rs
```diff
@@ -32,7 +32,7 @@ pub async fn create_test_environment(
             port: 5432,
             user: "clementine".to_string().into(),
             password: "clementine".to_string().into(),
-            dbname: format!("clementine_tx_sender_{}", get_current_test_name()),
```
aoengin, line 35: why did we change that?

atacann: db names become too long, other than that no particular reason I think

PR #1264, opened 2026-03-19
Add tracking to txsender

atacann:
- Adds tracking functions and rpc to txsender
- Change TX_SENDER_INCLUDE_UNSAFE to false in example env file


crates/tx-sender-types/src/lib.rs
```diff
@@ -13,3 +14,7 @@ pub use clementine::{
     ActivatedWithTxid, FeePayingType, InsertTryToSendParams, RbfSigningInfo, RbfSigningSpendPath,
     TxMetadata,
 };
+pub use tracking::{
+    ActivationBlocker, ActivationBlockerReason, ActivationState, BitcoinTxStatus, CommitRevealKind,
+    CommitRevealStatus, RevealStatus, TrackRequest, TrackResponse, TrackStatus, TxStatus,
```
aoengin: we can move Citrea related stuff under "citrea" feature flag

crates/clementine-tx-sender/src/tracking.rs
```diff
@@ -0,0 +1,513 @@
...
+
+        for row in rows {
+            if matches!(row.transaction_kind, TransactionKind::Aggregate) {
+                aggregate_finalized = row.aggregate_finalized;
+                if let Some(outpoint) = row.commit_outpoint {
+                    aggregate_commit_tx = Some(ctx.bitcoin_tx_status(outpoint.txid).await?);
+                }
+                if let Some(try_to_send_id) = row.try_to_send_id {
+                    let mut aggregate_submission = self
+                        .build_submission_status_by_id(ctx, try_to_send_id as u32)
```
aoengin: I know can't be negative but still safer to use try_from to prevent confusing errors

crates/clementine-tx-sender/src/tracking.rs
```diff
@@ -0,0 +1,513 @@
...
+            self.current_tip_height
+                .saturating_add(1)
+                .saturating_sub(height)
+                >= self.tracker.finality_depth
+        })
+    }
+}
+
+impl TxSenderTracker {
+    pub async fn track_request(&self, request: TrackRequest) -> Result<TrackResponse, BridgeError> {
```
aoengin, line 126: Some tracking info can be stale because it reads from the DB instead of checking RPC directly. For example, build_activation_state only uses DB data. This data is updated by the txsender loop, so if TX_SENDER_POLL_DELAY_MS is high, the tracking result can be outdated for longer.

crates/clementine-tx-sender/src/tracking.rs
```diff
@@ -0,0 +1,513 @@
...
+            let mut fee_payer_tx_info = ctx.bitcoin_tx_status(fee_payer.txid).await?;
+            fee_payer_tx_info.mined_at_height = fee_payer_tx_info
+                .mined_at_height
+                .or(fee_payer.mined_at_height);
+            fee_payer_tx_infos.push(fee_payer_tx_info);
+        }
+
+        let activation = self.build_activation_state(ctx, try_to_send_id).await?;
+
+        let status = if row.input_unspent_timed_out {
```
aoengin: if both input_unspent_timed_out and mined_at_height can both be set at the same time, the current precedence is wrong.

.github/workflows/release.yml
```diff
@@ -137,6 +137,9 @@ jobs:
             - clementine-cli-${{ github.ref_name }}-no-automation-linux-amd64
             - clementine-cli-${{ github.ref_name }}-with-automation-linux-amd64
 
+  docker_release:
```
aoengin, line 138: This can still pass even if the release job fails, so we can add release as a dependency to be safe.

atacann: it takes too long to publish with a dependency. Releases are manual anyway, they should build

crates/clementine-tx-sender/src/tracking.rs
atacann: For some  cases it makes sense to read from DB, especially if it determines if a tx is sent by txsender like build activations state. One purpose of these tracking types is so that we can debug without looking at DB if there is a problematic tx.

